PR #24, opened 2015-02-16
Added commonIdType

pstenbjorn:
@jungshadow I added commonIdType element and implementation.  Also following proposed guidelines in managing feature changes.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -174,7 +174,7 @@
               <xs:element name="name" type="xs:string" />
               <xs:element minOccurs="0" name="type" type="xs:string" />
               <xs:element minOccurs="0" name="number" type="xs:integer" />
-              <xs:element minOccurs="0"	name="common_id" type="commonIdType" />
+              <xs:element minOccurs="0" name="common_id" type="commonIdType" />
               <xs:element minOccurs="0" name="parent_id" type="xs:string" />
             </xs:sequence>
             <xs:attribute name="id" type="xs:string" use="required" />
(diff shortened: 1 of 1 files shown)


vip_spec.xsd
```diff
@@ -407,6 +414,17 @@
       <xs:enumeration value="NO"/>
     </xs:restriction>
   </xs:simpleType> 
+  
+  <xs:simpleType name="commonIdEnum">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="ocdid"/>
+      <xs:enumeration value="fips"/>
+      <xs:enumeration value="OCDID"/>
```
jungshadow: Do we need multiple cases for ocdid (e.g. caps and lower)?

pstenbjorn: I would think not. And probably best if we don't now that I think of it.

vip_spec.xsd
```diff
@@ -174,7 +174,8 @@
               <xs:element name="name" type="xs:string" />
               <xs:element minOccurs="0" name="type" type="xs:string" />
               <xs:element minOccurs="0" name="number" type="xs:integer" />
-			  <xs:element minOccurs="0"	name="common_id" type="xs:string" />
+              <xs:element minOccurs="0"	name="common_id" type="commonIdType" />
```
jungshadow: Not that this is a big deal, but is there a tab in between these attributes?

pstenbjorn: There is.  I replaced it in my local instance and committed it to this branch

jungshadow:
Perfect.

jungshadow:
@pstenbjorn Hmm...I think I merged too soon. I failed to notice that you hadn't removed the caps enumerations, yet. Want me to revert?

pstenbjorn:
Yes please.

Let me know when it is reverted and I will recommit.

PR #25, opened 2015-02-16
Revert "Added commonIdType"

jungshadow:
Reverts votinginfoproject/vip-specification#24


pstenbjorn:
Committed without caps

Issue #21, opened 2014-10-10
Structured hours for open and close times


jungshadow:
On any element that this applies (e.g. polling_location, early_vote_site), this will allow officials to provide machine-readable hours to data consumers. However, most location hours are largely unstructured partially due to the intake mechanism (i.e. election system) and the hours themselves (i.e. some are by appointment only or vary greatly from day to day). A few related questions:

* Are there states that designate location hours at the state level and do those rules hold for municipal elections, as well?
* In lieu of state regulations, are there simple systems that would allow counties to manage the locations and hours, reduce administrative burden, and provide data in a more structured format?

High level of difficulty to tackle, but worthwhile if it can be accomplished.

jktomer:
Lessons learned from collecting structured polling hours (independently of VIP) for early voting in 2014:

* Time zone information definitely needs to be specified explicitly per polling location. Otherwise you run into painful cases like Phenix City, AL -- whose municipal government runs on Eastern Time despite the fact that AL is officially in Central Time -- where it's impossible to correctly interpret any hours that are just specified as "local time."

* Because the fall DST boundary often occurs during the early voting period, time zones further should be specified in a way that accounts for DST. Suggested formats: 
  + ISO 8601 ("08:00-05:00" for 8AM Eastern Standard Time), attached to each individual time specification. Pro: precise and easy to interpret and validate without reference to any external database. Con: somewhat obscure; repetitive (since zone info has to be attached everywhere a time is specified); requires breaking entries for repeated daily schedules up at the DST boundary, if applicable
  + tzdb syntactic ("EST5EDT"). Pro: can be specified once for an entire location. Fairly easy to read and write. Con: untrained users may still find it pretty opaque. Hard to validate (to distinguish right from wrong values requires the tzdb, so it can't be done in an xml schema validator).
  + tzdb symbolic ("America/New_York", "America/Los_Angeles", "America/Metlakatla"). Pro: can be specified once for an entire location. Easy to interpret, even for untrained readers. Con: easy to mess up when writing, hard to validate.

[...]

PR #31, opened 2015-03-04
Polling Place Schedule

pstenbjorn:
Including example xml

@jungshadow I created a micro-schema for just the polling_location and a schedule element based on the comments of others.  There is also a valid sample xml doc for this schema.

### DIFF
--- a/pollplace_sample.xml
+++ b/pollplace_sample.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0"?>
 <vip_object schemaVersion="3">
   <source id="1">
     <name>Department of Elections, Commonwealth of Virginia</name>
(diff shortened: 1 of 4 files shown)


jungshadow:
This looks pretty good and incorporates some of @jktomer's concerns, but it doesn't seem to handle @pkoms concern in #21 specifically with "by appointment" (NB: that may be easily handled by a boolean in the schedule element). Is it redundant to have the timezone offset in both the date and as an independent element?

pstenbjorn:
@jungshadow there are a number of ways that @pkoms by appointment can be handled.  Would using the string value in `pollingLocationScheduleType.polling_hours` be sufficient?  My thinking is that if the polling_hours element is used then the schedule cannot be used programmatically.

I will be interested to see other thoughts on timezone.  In the XML example, I wanted to show two ways of doing it.  The `xs:date` datatype allows timezone, but it is optional.

chip-rosenthal:
Why is the schedule burst into so many (five) fields. Why isn't it just an opens and closes datetime?

pkoms:
I believe the 5.0 polling_location is meant to encapsulate early voting sites as well, which can't be represented as simply. (Since they don't stay open for 24 hours at a time.) n.b. a simple boolean wouldn't be sufficient for appointments -- appointments may apply only on specified days or within specified times. I think this is a step in the right direction, but doesn't give us all of the options we'd need.

jungshadow:
@chip-rosenthal Meant to write back to you on #21 today. First, thanks for staying engaged in this work.

In a lot of states, a simple open close datetime would be fine, but we've encountered a number of situations where there are gaps in the hours of operations (e.g. open from 8-10, closed until 2 PM, open from 2-4:30 PM). There are also situations where the hours are by appointment (as @pkoms just mentioned).

jungshadow:
@pkoms I agree and disagree with you :) Since the schedules are granular and can specify certain hours, why couldn't a particular schedule dictate that "between these hours, early voting is by appointment only," assuming there's a boolean that specifies "by appointment?" It overloads the definition of "open/close time," but maybe the elements be more appropriately named start/end (of a schedule).

chip-rosenthal:
@jungshadow  - What's wrong with this?
<schedule>
<opens>2015-01-01T08:00:00-06:00</opens>
<closes>2015-01-01T10:00:00-06:00</closes>
</schedule>
<schedule>
<opens>2015-01-01T14:00:00-06:00</opens>
<closes>2015-01-01T16:30:00-06:00</closes>
</schedule>

pstenbjorn:
Added a `by_appointment` boolean element to `pollingLocationScheduleType`.  

@chip-rosenthal and @jungshadow  I see nothing is wrong with expressing the full ISO 8601 date, time and timezone.  The most significant limitation now is that there are few source systems that give rise to these data.

jungshadow:
@pstenbjorn Could the full ISO datetime+tz be assembled from the existing data? What's the difficulty level? My guess is the tz data may be the most difficult thing to capture, especially in states that have more than one tz.

jdmgoogle:
Y'all are moving too fast for me (still trying to set up, clone, branch, etc, my git repo from github) so as a result you're getting this monstrosity of cut-and-paste. You brought it upon yourselves. :)

Here's our proposal. The patterns for the time elements are wrong, but ultimately they're going to be the regexp for full ISO time+TZ.


<xs:element name="hours_open">
  <xs:complexType>
    <xs:element type="xs:integer" name="hours_open_id" />
    <xs:element maxOccurs="unbounded" name="schedule">
      <xs:complexType>
        <xs:sequence>
          <xs:element type="xs:date" name="start_date" />
          <xs:element type="xs:date" name="end_date" />
          <xs:element maxOccurs="unbounded" name="hours">
            <xs:complexType>
              <xs:sequence>
                <xs:element maxOccurs="1" type="xs:time" name="start_time" pattern="HH:mm:ss('Z'|('+'|'-')HH:mm)" />
                <xs:element maxOccurs="1" type="xs:time" name="end_time" pattern="HH:mm:ss[('Z'|('+'|'-')HH:mm)" />
              </xs:sequence>
            </xs:complexType>
          </xs:element>
          <xs:element maxOccurs="1" type="xs:boolean" name="or_by_appointment" />
        </xs:sequence>
      </xs:complexType>
    </xs:element>
  </xs:complexType>
</xs:element>

jdmgoogle:
Aaaaand it ate all the formatting. Please hold.

pkoms:
@jungshadow Right, now I understand your recommendation better. That works for the individual intervals. Do you have thoughts on how to represent that a location is open by appointment *always*? If you have a location that's open from M-F 12-2, but by appointment any other hour, it adds a lot of extra lines to representing those hours.

Not sure we can improve on it, but I wanted to flag it in case you had an idea.

jungshadow:
@pkoms @pstenbjorn I just pushed a new branch with an [alternate schedule idea](https://github.com/votinginfoproject/vip-specification/blob/feature/alt-schedule/schedule.xsd) c/o @jdmgoogle. Can you both take a look at it and let me know what you think? There's a working example in the comments.

jdmgoogle:
Thank you, Jared. I'll get the hang of github and the git command-line soon. :)

The pattern for the start and end times would actually be:

(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]|(24:00:00))(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))

but represents the HH:mm, etc, shorthand currently in the file.

pstenbjorn:
In general this is great.  The hours as a child of schedule is a great improvement.  We will have to see about conforming existing data to the time patterns.  The challenge of inferring timezone can be managed if we can parse and geocode the polling location address.  There are lots of services to identify timezone by lat long.

jdmgoogle:
Paul,

Thanks. Two challenges of relying on geocoding to infer time zones, though, are (1) assuming that the geocoding service is correct for all the addresses, and (2) places like Phenix City, AL, which are technically in Central Time but everyone (including the government) use Eastern Time.

http://en.wikipedia.org/wiki/Phenix_City,_Alabama#Time_zone

pstenbjorn:
Combined proposed schema by @jdmgoogle into the pollplace.xsd (with minor modification).  This currently is valid and I will try to generate some sample xml once others can review:

See [the latest commit](https://github.com/votinginfoproject/vip-specification/commit/3e45ed8f81b471dc038a8023b81930a6469d0e8f)

pstenbjorn:
@jdmgoogle it sounds like we will need a time zone element in the schema.  The only usecase for it that I can think of right now is for polling locations, but in the future it may apply to results so should it be defined at the locality level and then inherited by related elements?  Or should it be defined within the elements that use the time zone?

pstenbjorn:
@jdmgoogle thanks for catching that, the `polling_hours` should be in the `pollingLocationType` element as an alternative to having a schedule.

pstenbjorn:
Published a sample xml for the updated schema.  Please take a look at [the sample xml](https://github.com/votinginfoproject/vip-specification/blob/feature/schedule/pollplace_sample.xml) and let me know if this matches your expectations.

pstenbjorn:
Updated the schema [in this commit] (https://github.com/votinginfoproject/vip-specification/commit/42c458dcb9b13fcefe292976d924b95c384f8f3b) and sample to match @jdmgoogle structure. `hours_open` element is now separate and related to `polling_location`, removed `polling_hours` element.  Please let me know what your thoughts are and we can consider putting this schema change into [the working draft schema](https://github.com/votinginfoproject/vip-specification/blob/vip5/vip_spec.xsd).

pollplace_sample.xml
```diff
@@ -0,0 +1,72 @@
...
+    <description>The Department of Elections is the official source</description>
+    <organization_url>http://www.elections.virginia.gov/</organization_url>
+    <feed_contact_id></feed_contact_id>
+    <tou_url></tou_url>
+  </source>
+  <polling_location id="80011">
+    <address>
+      <location_name>Appalachian College of Pharmacy</location_name>
+      <line1>1060 Dragon Rd</line1>
+      <line2></line2>
```
jungshadow: Sort of a nitpicky thing, but I'd prefer empty or null fields be absent. To me, it seems cleaner and I'd rather ensure we're providing best practices in our example files.

jungshadow:
I think this looks good. I'll echo @pstenbjorn's concerns that we'll simply lose data because the original data are stored as strings with no defined format. That said, I'm fine with introducing some rigor. It has my blessing, but I'd like @nomadaisy and @pkoms to weigh in on this before merging.

pstenbjorn:
Before we merge this into the [vip5](https://github.com/votinginfoproject/vip-specification/tree/vip5) branch, maybe we should merge these changes into the vip_spec.xsd?

I will make sure when we generate a sample xml next no empty tags are displayed.

pollplace_sample.xml
pstenbjorn: I will change my script that generates this to treat empty strings as null.  This is something I meant to do a while ago.

jungshadow:
@pstenbjorn Completely agree. @nomadaisy and @MariaDemocracyWorks are going to look at this today, but @pkoms won't be able to until Tuesday. Let's get this branch ready for a merge and, either, move to another discussion topic or try to tick off some of the working items. How does that sound?

pstenbjorn:
@jungshadow sounds perfect.  I will add the vip_spec.xsd with the additions and a sample xml using the schema changes.

pstenbjorn:
@jungshadow and others.  I have merged the schedule changes into the [vip_spec.xsd] (https://github.com/votinginfoproject/vip-specification/blob/feature/schedule/vip_spec.xsd) file and added [sample_feed_for_vip_spec.xml] (https://github.com/votinginfoproject/vip-specification/blob/feature/schedule/sample_feed_for_vip_spec.xml) which uses the scheduling elements in two ways.  One to represent an early voting site associated with locality and one to represent a polling place schedule on election day.  Since we had added the `common_id` element to `electoral_district` the sample also features the use of ocdids with electoral districts.  Please bear in mind that this is an incomplete feed and is not based on actual precinct and early voting assignment (VA does not have early voting).

nomadaisy:
I'm good with this structure.

jdmgoogle:
FYI @nomadaisy, @pkoms, @kep, @MariaDemocracyWorks, @jen-tolentino, @kennethmbennett, @cjerdonek

This coming Thursday, March 12th, @jungshadow, @pstenbjorn, and I will be meeting at Google to have a mini "hackathon" to try and come up with a release candidate version of the spec. Anyone who's in DC is welcome to attend in person, and anyone who isn't in DC but would like to VC in, please ping me and I'll make sure you're looped in.

For those who can't do VC or attend in person, no fear; all of our work product, etc, will still be published and tracked via GitHub, but having in-person talks will reduce the latency.

I've tried to Cc the relevant interested parties, but if I've left someone out please either feel free to Cc them yourselves or let me know who I should include.

Thanks.

kennethmbennett:
@jdmgoogle do you know what time the hackathon starts (Pacific time)?  I would like @Josh-LACRRCC  to participate.

+ @jungshadow @demcg

jdmgoogle:
The meeting will start around 2:30pm EDT/11:30am PDT.

pkoms:
In response to the previous request for comment: LGTM.

cjerdonek:
@jdmgoogle I'd like to hop on when it starts.  Can you e-mail me the info?  @jungshadow has my contact info.  Thanks.

jdmgoogle:
Definitely. It'll be via Google Hangout, so either I'll send you an invite directly once the meeting has started (if Hangouts works on your computer) or can dial you in to the meeting (if phone would work better).

cjerdonek:
Hangout is fine, thanks!

PR #32, opened 2015-03-05
Generic hours_open structure and sample use

jdmgoogle:
Here's my take on the polling place hours idea. It's actually more generic than polling place hours, in that something like an election administration office could define hours of operation as well, and use the same hours element by reference.


jungshadow:
Is this already merged into #31? If so, we can close this request and delete the branch.

jdmgoogle:
Yes, this has (largely) been merged into #31. Dropping this branch now.

PR #33, opened 2015-03-06
Feature/boolean cleanup

jdmgoogle:
The yesNoEnum was really just a way to do xs:boolean, so let's just use xs:boolean.


pstenbjorn:
@jdmgoogle if @jungshadow agrees I will make this change in the vip_spec.xsd in [feature/schedule branch] (https://github.com/votinginfoproject/vip-specification/tree/feature/schedule) so we can have one merge.

jungshadow:
I'm definitely fine with this change, but we may want to merge in some of the recent doc changes from the master branch before editing `index.html`. @nomadaisy is also going to make a massive update to the documentation that includes examples of the flat file format (NB: probably goes without saying, but a good chunk of the docs will have to be rewritten for the new spec).

jungshadow:
@pstenbjorn Actually, since it's a separate change, let's keep it clean and as a separate pull request. We should also open an issue on it if it doesn't have one. Cool? + @jdmgoogle

pstenbjorn:
@jungshadow I am happy to keep pull requests feature specific.  I will make the changes to the vip_schema.xsd file in this branch.

jdmgoogle:
I'm dropping this PR since this cleanup really needs to be merged into the existing big change instead of vip5 (what is the vip5 branch being used for right now, BTW?).

I can resubmit this PR to pull it into a different branch; just let me know which one.

jungshadow:
@jdmgoogle You got it right. vip5 is the dev branch for the spec. We'll eventually merge into master and tag it, but, for now, all changes go here.

jdmgoogle:
Okay. I somehow managed to bork my local repository trying to merge in the changes that have been made to this branch. I have no idea which version I actually have in my local repo right now, because it doesn't look like anything in the vip5, master, or my local feature/boolean-cleanup branch. It appears that several other large cleanups got committed to this branch? (Either that or I'm reading the github history incorrectly, which is equally likely.)

I followed the "Use the command line to resolve conflicts before continuing" instructions, which apparently is code for "dump random code into your repository and GOOD LUCK WE'RE ALL COUNTING ON YOU."

TL;DR: Someone else can have fun merging this because I messed up and I'm going to break it horribly.

jdmgoogle:
Okay, I think a "salt-the-earth" approach worked and this is ready for inclusion. My apologies that y'all have to see my Git/GitHub teething paints.

jungshadow:
@jdmgoogle Heh...we all went through it and those scars will never heal :)

This looks good to me. On a related note, I'm going to fill out another issue for us to nuke the `vip_spec_v5.0.xml` file, since we're not really using it and it's probably just causing more confusion than it's worth. You ok with that, @pstenbjorn?

cjerdonek:
> On a related note, I'm going to fill out another issue for us to nuke the vip_spec_v5.0.xml file, since we're not really using it and it's probably just causing more confusion than it's worth.

:+1: from me because it definitely caused me confusion.  I would actually love to see all the `vip_spec_*.xsd` files removed from the repo (as well as `sample_feed_for_*.xml`), so that the correct file for a version was instead located at `vip_spec.xsd` of the branch of that version name (and which would be the only spec file).  As of now, it's a bit confusing to see copies of all of the versions in all the branches.

pstenbjorn:
@jdmgoogle and @cjerdonek I am fine with nuking the `vip_spec_v5.0.xsd` and while we are at it we should remove the `sample_feed_for_v5.0.xml` file too.

PR #34, opened 2015-03-06
Proposal on how to handle external identifiers

jdmgoogle:
In several instances we've struggled to figure out the proper identifier to use for objects such as localities, precincts, etc. Here's our proposal on how to handle this. In short, the IDs we use to link the XML entities together could be xs:integer or even xs:ID (for better convergence with 1622) and then we use repeated external_identifier fields to augment the data. In the comments in the new file I've provided an example of how this could be used to provide additional information for a county (my old home of Durham County, NC) and Senator Al Franken.

### DIFF
--- a/external_identifiers.xsd
+++ b/external_identifiers.xsd
@@ -10,7 +10,7 @@
       <value>37063</value>
      </external_identifier>
      <external_identifier>
-      <type>GNIS</type>
+      <othertype>GNIS</othertype>
       <value>1008550</value>
      </external_identifier>
      <external_identifer>
@@ -40,7 +40,15 @@
 <xsd:complexType name="external_identifier">
   <xsd:all>
     <!-- SURVEY: Should 'type' be a string or an enumeration? -->
-    <xsd:element type="xsd:string" name="type"/>
+    <xsd:element type="identifierType" name="type"/>
+    <xsd:element type="xsd:string" name="othertype"/>
     <xsd:element type="xsd:string" name="value"/>
   </xsd:all>
 </xsd:complexType>
+
+<xsd:simpleType name="identifierType">
+  <xsd:restriction base="xsd:string">
+    <xsd:enumeration value="ocd-id"/>
+    <xsd:enumeration value="FIPS"/>
+  </xsd:restriction>
+</xsd:simpleType>
--- a/external_identifiers.xsd
+++ b/external_identifiers.xsd
@@ -10,30 +10,36 @@
       <value>37063</value>
      </external_identifier>
      <external_identifier>
+      <type>OTHER</type>
       <othertype>GNIS</othertype>
       <value>1008550</value>
      </external_identifier>
      <external_identifer>
-      <type>census</type>
+      <type>OTHER</type>
+      <othertype>census</othertype>
       <value>99063</value>
      </external_identifier>
 
      Here are a few for Senator Al Franken:
 
      <external_identifier>
-      <type>opensecrets</type>
+      <type>OTHER</type>
+      <othertype>opensecrets</othertype>
       <value>N00029016</value>
      </external_identifier>
      <external_identifier>
-      <type>votesmart</type>
+      <type>OTHER</type>
+      <othertype>votesmart</othertype>
       <value>108924</value>
      </external_identifier>
      <external_identifier>
-      <type>thomas</type>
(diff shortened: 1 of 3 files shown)


pstenbjorn:
@jdmgoogle I like this handling of common identifiers better than a commonIdType, it offers greater reusability.  I think it addresses my concern that identifiers like ocdid and FIPS can be used for both districts and geographic units.  @jungshadow if you want to merge this pull request, I can work on incorporating this change into the draft schema.

jungshadow:
@jdmgoogle @pstenbjorn I, too, like this, but, in taking the "comment survey," it might be useful to make `external_identifier.type` an enumeration. I can see the downsides of maintaining a possibly ever-growing list, though (e.g. the example notes "imdb" as a potential source). I'm fine with this being decided by majority rule.

pstenbjorn:
In spite of the heft of the enums, I think these need to be enumerated.

jungshadow:
@pstenbjorn Here's the crux of it: on the `external_identifier.value`-side, are we liable to supply and maintain what the standard value for a particular enum should be (e.g. the entire permalink for an imdb page or just the unique identifier...assuming people know how to find that)? Unfortunately, I think the answer is "yes."

pstenbjorn:
@jungshadow I think for the schema purposes for the `external_identifier.value` we are liable for the discreet value that is an identifier.  If we have enumerations there is liability for the entire permalink structure, that brings up another prospective schema issue.  While the distinct value for Al Franken's imdb entry would be 'nm0291253' the imdb permalink value would be http://www.imdb.com/name/.  Not sure if this is stored as one value or two separate values.  One without the other seems to have little value in any use case.  Let me lay out what one would think a reasonable subset of enumerations would look like and then we can debate whether these values will be free text or enumerations.

jdmgoogle:
I've updated the proposal to include support for both known "common" enumerated types and "other" types. I don't think we should be in the business of knowing the correct formatting for every single possible external identifier out there; part of the benefit of this approach is that Random State can include their own identifiers in the feed and have them propagate up, and we don't have to know about it. There are some common types about which we should know (e.g., FIPS, OCD-ID, etc) and we should curate those. But otherwise let it happen organically.

jungshadow:
@jdmgoogle In concept, I'm good with this. Do we need an `xsd:choice` surrounding the `type` and `othertype` elements? As it's currently set up, is it possible to have a valid `external_identifier` with all three subelements present?

Minimal issues aside, what do you think, @pstenbjorn?

jdmgoogle:
@jungshadow and @pstenbjorn, I believe the idea we kicked around (or at least the one that I liked) was to include an "OTHER" in the set of enumerated values, and we would require feed creators to use OTHER if they filled out the otherType field.

pstenbjorn:
@jdmgoogle I liked the idea as well.  I think your representative schema will work, with modifications below, with the addition of "other" in the enums.  @jungshadow, yes I think these elements should be wrapped in a `choice` container.  At some point today, I will edit the schema, nest it in the larger schema and generate some xml to do validation.

jdmgoogle:
I've updated the schema proposal to include OTHER type support. Please take another look.

jungshadow:
@jdmgoogle Still looks good :) I'm still wondering if there should be an XSD constraint that invalidates combinations of `type` (not "OTHER"), `othertype`, and `value` all being present.

pstenbjorn:
@jungshadow I will be happy to lay out the schema for a choice between `othertype` and `value`.

cjerdonek:
With the latest PR, what would be the line you'd need to add to the schema to support external identifiers for a particular entity?  Would it be something like the following?
```xml
<xs:element name="external_id" type="externalIdentifier" minOccurs="0" maxOccurs="unbounded"/>
```
Would it help to add a collection type so this can be something like the following?
```xml
<xs:element name="external_ids" type="externalIdentifiers"/>
```

pstenbjorn:
I committed the `vip_spec.xsd` file to this branch with this element included. The consistent way of structuring this was to add an element for `external_identifier` and include an attribute `id` that can be referenced by `state`, `locality`, `precinct` and `electoral_district`.  

NB: This has not incorporated `xs:ID` types.  That will be a in a later commit.  

This file has cumulative changes for schedule and external-identifiers branches.

Issue #42, opened 2015-03-10
Add Office object


cjerdonek:
Currently, the Contest object references the office via an "office" string field.  This issue is to suggest adding an Office object, in which case the contest field would become an office ID.

The advantage of doing this is mostly around DRY.  For example, if we add translations of the office name per issue #39 (which could potentially be many), it would be annoying to have to repeat these same translations in every Contest (e.g. for every district).  We may also find that offices have other useful attributes (e.g. the URL to the office's web site, physical address of the office, term of the office, description, etc).

Another advantage of breaking this out is that it's likely that the Office objects will remain the same from election to election, so Departments of Election could prepare this data separately (potentially by hand), independent of any particular election, and then reuse the same data.

djbridges:
I think this is a good idea, both on the merits Chris mentions and
(selfishly) because it will make it much easier for ballot information to
play nice with elected official/other civic engagement datasets.

On Tue, Mar 10, 2015 at 12:41 PM, Chris Jerdonek <notifications@github.com>
wrote:

> Currently, the Contest object references the office via an "office" string
> field. This issue is to suggest adding an Office object, in which case the
> contest field would become an office ID.
>
> The advantage of doing this is mostly around DRY. For example, if we add
> translations of the office name per issue #39
> <https://github.com/votinginfoproject/vip-specification/issues/39> (which
> could potentially be many), it would be annoying to have to repeat these
> same translations in every Contest (e.g. for every district). We may also
> find that offices have other useful attributes (e.g. the URL to the
> office's web site).
>
> Another advantage of breaking this out is that it's likely that the Office
> objects will remain the same from election to election, so Departments of
> Election could prepare this data separately (potentially by hand),
> independent of any particular election, and then reuse the same data.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/votinginfoproject/vip-specification/issues/42>.
>



-- 
Donny Bridges | Director of Civic Data | The Center for Technology & Civic
Life | 202.549.3717 | donny@techandciviclife.org

jdmgoogle:
Please see #43.

cjerdonek:
I've changed my thinking on this issue a bit since last week and after having a chance to think more and play around with examples.

[...]

PR #43, opened 2015-03-10
Port current/proposed ballot features from 1622.2

jdmgoogle:
...to VIP.


cjerdonek:
Thanks, @jdmgoogle.  For the purposes of VIP, is there a requirement to match what VSSC 1622.2 has exactly, or is this just a starting point from which VIP can depart (e.g. if VSSC 1622.2's current iteration is lacking in any way)?  Also, if VSSC hasn't adopted this yet, maybe improvements developed here can flow back to them.

jdmgoogle:
My personal goal regarding ballot information is to settle on best practices and have both VIP and P1622.2 adopt them. I (and others) felt that the P1622.2 ballot model was more thorough and clean than the VIP model, hence the attempt to harmonize them. P1622 has a slightly larger group of people involved and will likely have more inertia, so-to-speak, but personally I would like to see the best possible solution adopted by VIP and P1622.2 in parallel.

pstenbjorn:
@jdmgoogle great start.  I will how this embeds in the VIP schema and give recommendation afterwards.

cjerdonek:
A note about feedback on this PR: this PR covers a bunch of broader areas, some of which were previously broken out into different topic issues (e.g. candidate statuses, vote variations, office object, etc).  Do we want to discuss all of these different issues in this single PR thread?

I fear it might get unwieldy in terms of length.  Also, there will likely be a bunch of different threads interwoven in the ordering, which will make it harder to follow.  One thought is to develop consensus on each individual topic area in a separate issue (while looking at this PR), and then update the PR in response afterwards.

jdmgoogle:
Normally I would agree with you, but the specific issue at hand here is "should we adopt the P1622.2 ballot information structure in VIP?" If there are portions of this PR which don't work or are problematic, then ultimately the root cause of the problem is with the P1622.2 spec. In that case those concerns need to be routed back to the 1622 committee. For the sake of simplicity -- i.e., not having to track 1622 issues across N different PRs -- I'd like to keep all 1622-related discussion in this PR.

cjerdonek:
If this PR is merged, will we still be able to improve upon and modify what was merged (and upstream changes to the committee afterwards), or are you saying that any changes would have to go through 1622 first?  If the latter, it sounds like VIP isn't the appropriate place to discuss portions of the spec covered by 1622.

jdmgoogle:
I wouldn't be opposed to small changes, but

a) they would need to be demonstrably better than what's currently in the spec; and
b) they should ultimately be merged back into P1622.2.

As I said before, the goal is to identify best practices for representing ballot information and then use that in both VIP and P1622.2.

cjerdonek:
Okay, I can see that I will have lots of feedback on this PR.  (Again, it seems like trying to discuss all of this in a single thread is way too much.  After all, this was the subject of an entire working group!)  If it's all or nothing with only small changes allowed afterwards, my personal opinion would probably be no.  I do want to identify best practices, but it sounds like you're saying we wouldn't be allowed to discuss "middle-sized" changes here, and that those discussions would need to go through the working group.

jdmgoogle:
Could you please explain what some of your concerns are about this model and how it's worse than the current VIP model?

jungshadow:
@cjerdonek Would you be open to reaching out to me and the VIP team at info@votinginfoproject.org? I think it would be great to chat and see if we might be able to understand your particular use case.

cjerdonek:
I'm not necessarily saying it's worse.  My main concern is with the condition or idea that after it's merged, we wouldn't be allowed to make significant changes without going through a non-public working group (i.e. it will be set in stone without their approval).  If I knew VIP could improve and do their own thing after merging, this concern would go away.  If VIP is more nimble and transparent than VSSC, it seems like it would be easier to make larger changes here and then communicate those back to them, rather than vice versa.

jdmgoogle:
VIP may be more nimble than P1622 -- and it's actually pretty straightforward to become an observer in the P1622 process -- but that's not the real problem. The real problem is that this data is stored in state and county Election Management Systems, and having multiple target data formats to which the states and counties would need to export this information would make it less likely that we would actually get the data. Having a single format increases the likelihood that we'll actually get the data.

If you have questions about the data, where it comes from, and how it gets exported to the VIP format (and to P1622 in the future) @pstenbjorn would be a good person to contact.

cjerdonek:
I'm well aware of those issues.

jdmgoogle:
In that case if you still have concerns about the actual content of the P1622 spec as we try to work it into VIP it would be really helpful for us to understand what those concerns or potential changes are. This thread would work, as would the info@votinginfoproject.org address @jungshadow mentioned. Alternately the latest P1622.2 revision should be out Real Soon Now. If you have questions about that then the person to contact is John Wack at NIST (john.wack@nist.gov).

cjerdonek:
Thanks, @jungshadow. I will reach out.

And @jdmgoogle, I'll share one of the more significant points of feedback, probably by tomorrow morning (relating to the office object).  And yes, I have been meaning to get involved in the working group (some others have also encouraged me to do so).  I will try to make that happen within the next week or so.

cjerdonek:
Here is some feedback re: the "Office" object in the PR.

I would like to request that the information about offices that isn't necessarily tied to a specific election be included in their own objects.  You could think of this as the "election-independent" office information (aka the offices for a _jurisdiction_ rather than for an election).

Doing this would help on the larger scale because (as I mentioned in issue #42) it would allow this information to be prepared once per jurisdiction and then reused more easily from election to election (possibly with minor updates or enhancements).  And because it would only need to be prepared once, there would be an incentive for jurisdictions or third parties to do this better and more completely (and manually if necessary, even without access to an EMS).  For example, translations of each text field could be added.

The current PR has an "Office" object with fields like electoral_district_id, name, termStartDate, and termEndDate.

In an election-independent formulation, you wouldn't have fields for term start and end dates, because those relate to a specific election.  Instead, I could see an "Office" object for things like Mayor, City Attorney, and Member of the Board of Supervisors, which contains things like name, description, URL, physical address, and length of term.  And to address the case of offices that differ only in a district number, for example, I could see a "Seat" object that would have an office ID, together with a district ID of some sort (or "at-large") to indicate the particular member of that office.

Alternatively, the "Office" object could contain the reference to the district, if it exists.  And for cases like "Member of the Board of Supervisors," the Office object could have a reference to a body ID (which in this case would reference "Board of Supervisors").

In both cases, VIP's Contest object could reference the office or seat ID.  Information like the start and end date of the term could perhaps go in the Contest object in the election in which that office is being filled.

cjerdonek:
Between the two options I described above of (1) not including the district number in the Office object (and having a Seat object containing the district) and (2) including the district number in the Office and having a Body object, I think I would lean towards the latter.

In the case of (2), the Office "name" field could perhaps even be made optional and obtained instead from the Body object when it exists.  This would be useful for offices that are members of a body as it would be more DRY (e.g. members of a School Board elected at large or members of a Board of Supervisors elected by district, as they would all have the same name).  In addition to a "name" field, a Body object could have an attribute like "office name."  For example, for a Board of Supervisors, the body name could be "Board of Supervisors," while the office name would be "Supervisor."

cjerdonek:
What is the status of this issue, by the way -- are we waiting to hear back from 1622.2 on anything, and who was responsible for doing that?  (I was also curious which VIP members are active in 1622 as either members or observers.)

My proposal for issue #42 (and action item from last week's meeting) is a simple tweak to this PR, namely to add an external identifiers collection to the Office object.  But I wanted to wait until after the PR was merged before creating that proposed PR.

jdmgoogle:
We have the tentative go-ahead from 1622.2, but John Wack is going to run the suggestions by a larger group and see how they comment. I believe the timeline for that is "the next few days" but I'll ping him Friday right after lunch to see if he's heard anything new.

I'm fully in support of adding an external identifiers element to the Office object.

cjerdonek:
> I'm fully in support of adding an external identifiers element to the Office object.

Oh, in reviewing the PR, I see it is actually already there. :grinning: Incidentally, I did notice that the Person, Candidate, and Party objects don't yet have it, which could be useful (e.g. if a jurisdiction's EMS or voter registration database has its own ID's for such objects).

pstenbjorn:
@jdmgoogle I have been looking at ways to incorporate the ballot/candidate/contest items into VIP and I have a few questions.  In your comments in the XSD you ask `<!-- Question: What's the use case for ballotSelectionID? -->` and in my interpretation of the 1622.2 schema, these elements are used to display election results for contests (and any element extended from contests) so I think we should consider if there is a use case for `BallotSelection` in VIP.  

Second, structurally your example defines only `complexType` objects.  Is your suggestion that we adopt a similar structure to the 1622.2.  In it they define a parent element that is of a complexType that relates all other complexTypes or should these be converted into elements?

jungshadow:
@jdmgoogle I'd love to get this merged in, but I think @pstenbjorn's [outstanding questions](https://github.com/votinginfoproject/vip-specification/pull/43#issuecomment-83721283) are keeping this from moving. Also, when the reference spec is solidified, are you fine with integrating it into `vip_spec.xsd` and make sure it validates?

jdmgoogle:
@jungshadow I'll grab the latest version of the 1622 spec and integrate any changes into this PR and re-send it out. I still haven't gotten a great answer from the 1622 people as to *how* ballotSelectionID will be used.

As for complexType versus simpleType, I'd recommend keeping the parent/child relationship so there's less duplication of fields.

I'll update this today and send it out. Tomorrow through April 5th, though, I'm going to be on vacation. I'll make sure this gets taken care of before I leave.

jdmgoogle:
I need to run off for the moment to take care of some other things, but I pushed an updated-and-mostly-cleaned-up version of the most recent 1622.2 ballot schema to the feature/p1622-ballots branch. I'll get back to the rest of this tonight, but please take some time to go over the ballot information and layouts. You can see the most recent 1622 spec (v38) here:

grouper.ieee.org/groups/1622/groups/2/index.html

Look for the "Version 38" line.

pstenbjorn:
@jdmgoogle thanks for sharing the revised 1622.2 schema.  I will digest it.  A very specific question to follow up on my previous question.  

You have many complexType objects like `party` below.

    <xs:complexType name="Party"> 
    <xs:sequence>
    <xs:element name="abbreviation" type="xs:string" minOccurs="0"/>
    <xs:element name="name" type="xs:string"/>
    <xs:element name="external_identifier" type="external_identifier" minOccurs="0" maxOccurs="unbounded"/>
    <xs:element name="colorCode" type="xs:string" pattern="[0-9a-f]{6}" minOccurs="0"/>
    <xs:element name="logoUrl" type="xs:anyURI" minOccurs="0"/>
    </xs:sequence>
    <xs:attribute name="id" type="xs:ID"/>
    </xs:complexType>

The party_id is referenced elsewhere not the Party type.  So is your suggestion that we create element instances for each of these complex types in the schema?  That would mean that we would have a series of elements that looked like this:

`<xs:element name="PartyInstance" type="Party" />`

Wouldn't it be simpler for us to declare elements that were complex types?  VSSC took a different approach.  They declared a single element instance `ElectionReport` and all other complexTypes are related to this primary element.  Just wondering what your thoughts on these different schema construction approaches are.

jdmgoogle:
@pstenbjorn Ultimately I'd like for the VIP declarations to look and behave like the rest of the VIP declarations, and 1622.2 has their own thing. I'd like to enable compatibility without requiring them to be 100% identical.

I've gone through and cleaned up a bunch of stuff (I believe) and changed them to be more in the VIP model.

Unfortunately I really do have to jump out of this discussion right now. I leave the final decision in your capable hands. I'll be back online on 4/4.

jungshadow:
@pstenbjorn Do you have the time/bandwidth to finish up this pull request? In some of our other conversations, I think some of these models are a dependency for additional work (e.g. `Person` and `Contact` should be extended/used by/for `election_official`). If you don't, let me know and I'll try to work it in.

pstenbjorn:
@jungshadow I will be able to merge this in.  I will strive to have it done by tomorrow (4/1).

pstenbjorn:
@jungshadow I committed https://github.com/votinginfoproject/vip-specification/commit/cec058aabe7a68c5fe56370510690d92a2350956 which added a `vip_spec_with1622.xsd` file since the vip_spec.xsd file in this branch is fairly out of date.  I can merge these changes into the vip_spec.xsd but I want to make sure we retain the markup history on that document.  This validates.  There are some nuances to how the xsd had to be refactored to validate that might give rise to complexities in generating an xml file.

Issue #39, opened 2015-03-07
Support providing text field data in multiple languages


cjerdonek:
For all text fields, the spec should have a way to allow providing that information in multiple languages.  In San Francisco, for example, it is a requirement that all election information be provided in English, Spanish, and Chinese (as well as in Filipino starting January 1, 2016).

Currently, it seems like VIP consumers wouldn't be able to meet the same language requirements that jurisdictions may have (unless there is a way of providing additional languages that isn't documented in the [web site documentation](http://votinginfoproject.github.io/vip-specification/)).

jungshadow:
I'm adding this to the "Version 5.0" bucket, but there may be some push back. Personally, I want this to happen and soon. It will ultimately come down to how easy it is to incorporate translations in every VR/EMS system (assuming the translation data is even held in either).

I imagine that @pstenbjorn might have some insight on this.

cjerdonek:
One of my action items from the meeting was to propose something for this issue.

Before coding it, I wanted to describe what I was going for.  My goal was to make adding multi-language support to an element in the schema as simple and DRY as possible, for example by changing this--
```xml
<xs:element name="greeting" type="xs:string"/>
```
to this--
```xml
<xs:element name="greeting" type="multiLangText"/>
```
A concrete example would look like this--
```xml
<greeting>
    <text lang="en">Hello</text>
    <text lang="es">Hola</text>
    <text lang="fr">Bonjour</text>
</greeting>
```
Ideally, the following would also be acceptable (if only English were available and for backwards compatibility, etc)--
```xml
<greeting>Hello</greeting>
```
Does this seem good to people?

cjerdonek:
And here is a rough stab at a schema definition for the idea described in the previous comment (I make no claims to have expertise in XML):

```xml
<xs:complexType name="langString">
    <xs:simpleContent>
        <xs:extension base="xs:string">
            <xs:attribute name="lang" type="xs:string" use="required"/>
        </xs:extension>
    </xs:simpleContent>
</xs:complexType>
<xs:complexType name="multiLangText">
    <xs:union>
        <xs:simpleType>
            <xs:restriction base="xs:string"/>
        </xs:simpleType>
        <xs:complexType>
            <xs:sequence>
                <xs:element name="text" type="langString" minOccurs="1"/>
            </xs:sequence>
        </xs:complexType>
    </xs:union>
</xs:complexType>
```

pstenbjorn:
[...]

PR #49, opened 2015-03-13
Add a type to support text in multiple languages

cjerdonek:
Here is an initial PR to address issue #39, which incorporates @pstenbjorn's suggestion to use `xs:language`.


jungshadow:
@cjerdonek Thanks for this (and other pull requests). Can you make this pull request against `vip_spec.xsd` in the vip5 branch? Master's the current, stable code and we're working off the vip5 branch for any dev work. Eventually, we'll merge into master and tag it as a release, but we'd rather get things to a more stable point.

cjerdonek:
> Can you make this pull request against vip_spec.xsd in the vip5 branch?

Sure thing.  Note that I may do it tomorrow though (want to think through another approach).

cjerdonek:
Replaced by pull request #52, per @jungshadow's request.

PR #55, opened 2015-03-17
Feature/gitattributes addition

### LINKED ISSUE #54: Add .gitattributes file to enforce LF endings
There are a number of files that inconsistently have CRLF or LF endings. This is to enforce a single convention.

jungshadow:
This fixes #54.


jungshadow:
@pstenbjorn @cjerdonek There's not much to review here (besides lots of line-ending changes), but if you can give me a quick thumbs up on this, that would be great.

pstenbjorn:
@jungshadow looks like it formatted them correctly.  Thanks

jungshadow:
@cjerdonek Hmm...the way I initially read it was that the first line only works for users that haven't set their global setting. The second line is the enforcement that overrides the user's preference. It looks like I may have misinterpreted it. I'll fix, update, and merge.

jungshadow:
@cjerdonek This first line in the [GitHub docs example](https://help.github.com/articles/dealing-with-line-endings/#example) on this worries me a bit based on what it insinuates:
```sh
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto
```
But I'll go with my interpretation of the git docs. It's not a hard fix, if we have to go back to it.

Issue #61, opened 2015-03-24
Restructure repository and delete superfluous files


jungshadow:
Minimize the number of repository files for subsequent releases. The repository should contain:
* `examples`
* `README.mdown`
* `index.html`
* `vip_spec.xsd`
* `sample_feed.xml`

Delete the following files:
* `sample feed for v2.1.xml`
* `sample feed for v2.2.xml`
* `sample feed for v2.3.xml`
* `sample feed for v3.0.xml`
* `sample_feed_for_v4.0.xml`
* `sample_feed_for_v5.0.xml`
* `spec_v2.2_documentation.html`
* `vip_spec_v2.3_documentation.html`
* `vip_spec_v3.0_documentation.html`
* `vip_spec_pending.xsd`
* `vip_spec_v2.1.xsd`
* `vip_spec_v2.2.xsd`
* `vip_spec_v2.3.xsd`
* `vip_spec_v3.0.xsd`
* `vip_spec_v4.0.xsd`
* `vip_spec_v4.1.xsd`
* `vip_spec_v5.0.xsd`

cjerdonek:
:+1: !

pkoms:
I'm almost completely behind this request, but are you sure you want to kill the 3.0 documentation when it will still be supported by Google?

jungshadow:
@pkoms I'd actually like to make it more accessible. I'm going to post one of my comments on PR #62 here. A few things I didn't outline in my thinking:

1. We tag the current master branch at the latest commit as v3.0 (after a couple of small changes).
    1. This will allow users to easily browse to that point in history.
    2. Additionally, GitHub will archive that commit as a downloadable zip/tar.
    3. We could also tag previous SHAs for other releases (e.g. 2.3, 2.1, etc...), as well.
2. We need to figure out a way to manage our documentation so that the versions are also accessible, not just the current release.
    1. I believe the Open Civic Data folks use Read the Docs, but it would significantly change the way we write/manage our docs.
    2. Jekyll is built into GitHub pages, but I'll need to do a bit more research on how that might help us in this case.

There's been a bit more discussion about it on the PR, but I definitely want to make sure the docs are accessible for the current release and (at least) one previous version.

pkoms:
Thanks for expanding on this -- that's what I get for catching up on email last-first. I'll move my comments to the other PR.

jungshadow:
Closed by #62

PR #62, opened 2015-03-24
Remove superfluous files and rename sample feed

jungshadow:
See #61


jungshadow:
@pstenbjorn @jdmgoogle @cjerdonek I essentially burned it down and salted the earth. This should remove any and all ambiguity. Let me know if I went too far :)

jdmgoogle:
I'm all for blowing away old and unused versions of the spec, but we should keep the v3.0 files since not everyone is going to switch over to 5.0 right away.

Going forward we should have a policy that the current GitHub page will always have at least 2 and at most 3 versions of the spec up: the deprecated one, the current stable one, and (optionally) the version under development. Technically this would mean that currently we should leave 2.3, but does anyone still use that version?

pstenbjorn:
@jungshadow I tend to agree with @jdmgoogle that supported versions should be included in the repo and the most recently deprecated version would also be reasonable.  I agree with renaming the sample feed.  Is there any benefit to moving these items to a vip-archive repo?  For posterity sake if no other reason.

jungshadow:
@jdmgoogle @pstenbjorn A few things I didn't outline in my thinking:

1. We tag the current master branch at the latest commit as v3.0 (after a couple of small changes).
    1. This will allow users to easily browse to that point in history.
    2. Additionally, GitHub will archive that commit as a downloadable zip/tar.
    3. We could also tag previous SHAs for other releases (e.g. 2.3, 2.1, etc...), as well.
2. We need to figure out a way to manage our documentation so that the versions are also accessible, not just the current release.
    1. I believe the Open Civic Data folks use [Read the Docs](https://readthedocs.org/), but it would significantly change the way we write/manage our docs.
    2. Jekyll is built into GitHub pages, but I'll need to do a bit more research on how that might help us in this case.

I do have a method to my madness...contrary to popular opinion ;)

jdmgoogle:
I'm all for older versions being archived somewhere. My main concern is having the most-recently-deprecated, the current, and the in-progress versions of the spec being somewhere that's trivial for an election official or non-technical person to browse. If your approach means that a non-technical person could come here (having never seen the GitHub UI) and find v3.0 of the spec within 10-15 seconds, then I'm fine with that. If that's not the case then I argue we need a slightly different approach.

jungshadow:
@jdmgoogle I think the way that a non-technical person happens upon the spec is largely going to be through our outreach–a guided introduction, if you will–and the docs (do non-technical people actually browse GitHub?) ;) Additionally, in the docs, we should put links to the GitHub downloads, so no one actually has to interact with the GitHub UI. That said, I don't think the GitHub UI is that bad/hard.

cjerdonek:
I just want to chime in that I don't think the [personas](http://en.wikipedia.org/wiki/Persona_%28user_experience%29) of non-technical people or election officials should have any bearing on development workflow or structure of the repo.  That seems to me to be an orthogonal concern.  If there are certain needs there, those can almost certainly be handled with automated tooling that creates a public-facing view of what's in the repo and its branches, etc (e.g. providing static links to human-readable each of the versions on a web page).

Also, just to be clear, deleting the files from a branch doesn't mean the files won't be "removed from the repo." :grinning: They will still be there, but in another branch.  The "normal" approach for development projects is to have each version of something live in its own tagged branch (and not having copies of the different versions floating around).  Among other things, keeping copies around creates confusion because it's not clear which one is the "right" one (another case of where DRY helps).

By the way, on the subject of versions and public-facing info, when I first came to VIP's GitHub [page](http://votinginfoproject.github.io/vip-specification/), I would have liked to see documented, human-viewable pages for all of the versions.  As it is now, you only see v3 of the spec in the nicely-rendered form, and have to go to GitHub to see the others.  But again, this is orthogonal to development workflow.  Maybe to address people's concerns, the public-facing need could be addressed before deleting the files from the branch?

pkoms:
@jungshadow or @cjerdonek could you send me an example of human-viewable documentation you like that's powered off of GitHub files? My instincts are with Justin -- that we should care about the navigability of the relevant GitHub repos -- but I could be persuaded otherwise.

cjerdonek:
@pkoms Here is one (it's a project I started last week): http://cjerdonek.github.io/sf-base-election-data/

It's hosted on GitHub pages, and I have scripts to generate the HTML from files in a GitHub repo and push to GitHub pages.

The documentation for Python is a much more involved example: https://docs.python.org/3/

You can see that it has a pull-down in the upper-left corner that lets you toggle the version shown.  This documentation is generated using a program called Sphinx.

Read The Docs (which I believe @jungshadow mentioned) is a free solution that supports displaying multiple versions out of the box.  You can see an example of RTD in action here: https://docs.readthedocs.org/en/latest/index.html

The process for VIP could take any number of forms.  But it could be as simple as a shell script that copies HTML files from different branches (if we wanted to roll our own).

jungshadow:
@pkoms The [Open Civic Data Identifier Documentation](http://opencivicdata.readthedocs.org/en/latest/ocdids.html) is also hosted on Read the Docs. Clicking on latest also allow you to bring up the previous docs.
screen shot 2015-03-24 at 3 27 06 pm

jungshadow:
@pkoms @jdmgoogle @pstenbjorn Ideally, I agree with @cjerdonek that we should have a solution for the documentation first, but, since it's a rather involved change, I'd vote for simplifying the repository first and hacking away on the new docs as we go (_**NB:** this week is especially bad for me to make any decent headway_). I don't want to spend too many cycles on this issue, especially since we have more substantive issues to resolve (PR #43 for example) :)

cjerdonek:
Btw, an easy interim solution could be to create a branch called "archive" (which today would be the same as the current master), and then have a prominent sentence in the README directing people to that branch for earlier versions of the spec.  A fuller solution could then be worked out later.

jungshadow:
@cjerdonek The current master serves that purpose while we're developing on the vip5 branch. I'm hesitant to add additional branches. We have a large number of active branches right now, so using branches for storing versions/archives seems confusing. I'm planning to add a tag to the SHA of the current master and create a GitHub Release.

cjerdonek:
@jungshadow Yes, you're right. A tag would be better. In any case, my main point was that a link could be added to the README in the vip5 branch so that the text would be there when vip5 is merged to master. The link target could change once a tag is created, etc, without changing the appearance of the README.

jungshadow:
Added a tag for [the current state of the repository](https://github.com/votinginfoproject/vip-specification/releases/tag/v3-archive). I'm going to call this "done" and merge. I will wait 15 mins for any backlash ;)

PR #64, opened 2015-04-02
xs:ID and xs:IDREF cleanup

### LINKED ISSUE #44: Use xs:ID and xs:IDREF for identifiers
We should move away from using xsd:integer and/or xsd:string types as identifiers, and move towards a more strongly typed model using xsd:ID and xsd:IDREF in as many places as is reasonable. This will allow for more rigorous checking of data and types, and allow everyone along the processing chain to verify that references are (at least) to items of the correct type.

Comments? Feedback?

jungshadow:
Replaces the old string ids and string id references with xs:id and xs:IDREF, respectively.


jungshadow:
This will fix #44

jungshadow:
This also fixed #72

PR #65, opened 2015-04-02
Removed extraneous values in oebEnum

### LINKED ISSUE #46: odd_even_both enumeration values should be reduced to three choices
There are many string variants in the `oebEnum`. There should be only three (i.e. 'odd', 'even', 'both').

jungshadow:
Leaves only "odd", "even", and "both" as valid values.


jungshadow:
This will fix #46 when merged.

cjerdonek:
:+1:

PR #73, opened 2015-04-02
Creates a style guide

### LINKED ISSUE #41: Create a coding style guide
I'm wondering if we should follow the 1622.2 style guide. I mentioned this in a separate comment to @pstenbjorn a while back. If I were to break them out as a guess:
* attributes seem to be camel-cased (NB: exception being object_id),
* elements seem to be title-cased with no separators, and
* enumerations seem to be lower-cased with hyphens taking the place of spaces.

It would mean a lot of changes, but it would be nice to have an agreed-upon coding convention and a bit more similarity between the two specs, especially as we're planning to consume the ballot models. If everyone agrees, I'll create a pull request with the brief "style guide."

jungshadow:
Creating a pull request for @demcg's work. Fixes #41

### DIFF
--- a/StyleGuide.mdown
+++ b/StyleGuide.mdown
@@ -8,7 +8,7 @@
 
 ### File Specific Styles
 + All files should be encoded in utf-8.
-+ 4 Spaces should be used instead of tabs (or vice versa).
++ 2 Spaces should be used instead of tabs (or vice versa).
 + All trailing white space should be removed from the end of lines.
 
 ### Attribute Ordering
--- a/StyleGuide.mdown
+++ b/StyleGuide.mdown
@@ -1,0 +1,23 @@
+# Coding Styles
+
+## XSD Styles
+### Naming Styles
++ Attributes should be camelCased - The first letter of the second and subsequent words are capitalized.
++ Elements should be TitleCased (PascalCased) - The first letter of every word is capitalized.
++ Enumerations should be spinal-case - all lower case with spaces replaced with hyphens.
+
+### File Specific Styles
++ All files should be encoded in utf-8.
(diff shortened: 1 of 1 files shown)


StyleGuide.mdown
```diff
@@ -0,0 +1,23 @@
...
+
+## XSD Styles
+### Naming Styles
++ Attributes should be camelCased - The first letter of the second and subsequent words are capitalized.
++ Elements should be TitleCased (PascalCased) - The first letter of every word is capitalized.
++ Enumerations should be spinal-case - all lower case with spaces replaced with hyphens.
+
+### File Specific Styles
++ All files should be encoded in utf-8.
++ 4 Spaces should be used instead of tabs (or vice versa).
```
jungshadow: Personally, just because the nesting can get out of control, I'd rather have two spaces as opposed to four (no...I'm not trying to start a flame war, just aim for easier readability on the web, etc...)

StyleGuide.mdown
```diff
@@ -0,0 +1,23 @@
+# Coding Styles
+
+## XSD Styles
```
jungshadow: Though 1622 doesn't specifically have a style guide, I'd like to stick as close to their conventions as much as possible, even if it may be somewhat contrary to our own likings at times.

cjerdonek:
This looks good to me as a start (with the two-space change as @jungshadow mentioned).

cjerdonek:
Also, I think we should mention where particular rules are chosen for consistency with 1622 (e.g. by adding a parenthetical note to such bullet points), but we can add that later.

demcg:
I will make the 2 space change and merge the branch into VIP5

Thank you @jungshadow for opening this pull request.

Issue #81, opened 2015-04-03
convert sample_feed.xml to UTF-8


cjerdonek:
I noticed that sample_feed.xml is UTF-16.  This causes git to think it is a binary file.  I will post a PR to convert this file to UTF-8 (it is the only remaining file that is not UTF-8).

cjerdonek:
Addressed by PR #82.

PR #82, opened 2015-04-03
Address issue #81: convert sample_feed.xml to UTF-8

cjerdonek:
This addresses issue #81.


jungshadow:
@cjerdonek The diff doesn't work because the original is a binary file, but, hey, I trust you :)

cjerdonek:
@jungshadow Yep! I just "Saved As" in TextMate. :grinning:

Issue #83, opened 2015-04-03
Clarify that file styles apply to all files and not just XSD files


cjerdonek:
I noticed that the UTF-8 and trailing whitespace guidelines are in the XSD section, but those should really be about all text files.  I'll create a PR.

cjerdonek:
I created PR #84 for this.

cjerdonek:
Addressed by PR #84.

PR #84, opened 2015-04-03
Address issue #83: style for all text files.

cjerdonek:
This addresses issue #83.  The rendered form can be viewed [here](https://github.com/votinginfoproject/vip-specification/blob/feature/issue-83-file-styles/StyleGuide.mdown).


jungshadow:
LGTM

Issue #77, opened 2015-04-03
Change xs:string to internationalizedText as appropriate


cjerdonek:
This issue is to change "xs:string" to "internationalizedText" as appropriate.

This is something that I believe was done in one of the earlier PR's, but then got lost.  I can take care of this by reviewing the old comments and PR's, and then create a new PR.  I will wait until after the style stuff is done before making the PR.

cjerdonek:
I started working on this in branch [`feature/issue-77-internationalized-text`](https://github.com/votinginfoproject/vip-specification/tree/feature/issue-77-internationalized-text).

Also, here are the elements to update from the issue #39 comments:

ballot_response.txt:
* [ ] text

candidate.txt:
* [ ] party
* [ ] biography

contest.txt:
* [ ] primary party
* [ ] electorate_specifications
* [ ] office

custom_ballot.txt:
* [ ] heading

early_vote_site.txt:
* [ ] directions
* [ ] voter_services
* [ ] days_times_open

election.txt:
* [ ] registration_info
* [ ] absentee_ballot_info

election_official.txt:
* [ ] title

polling_location.txt:
* [ ] directions

referendum.txt:
* [ ] title
* [ ] subtitle
* [ ] brief
* [ ] text
* [ ] pro_statement
* [ ] con_statement
* [ ] passage_threshold
* [ ] effect_of_abstain

source.txt:
* [ ] description

election.txt:
* [ ] name
* [ ] election_type

party.txt:
* [ ] name

demcg:
@cjerdonek 

LAC uses a translation of the candidate's name as part of the ballot for non-latin character languages. I am not sure if this fits into the standard, or if it does where it fits.

![image](https://cloud.githubusercontent.com/assets/481571/7055547/105250da-ddf9-11e4-8f7d-366f258145fa.png)

cjerdonek:
@demcg Yes, I included Candidate.BallotName for that reason.  Thanks.  I'm in the process of typing up some thoughts about this issue and others.

cjerdonek:
Okay, I created PR #92 to address this.

I think the PR takes care of a lot of the easy and/or obvious stuff, but I think it would be good to start thinking about a more careful approach and rules around this going forward.

Here are some thoughts to consider around this, to give you an idea:

[...]

PR #92, opened 2015-04-08
Address issue #77: change xs:string to InternationalizedText

cjerdonek:
This is an initial attempt at issue #77.


jungshadow:
LGTM. Not letting "the perfect be the enemy of the good," I think we should accept these changes and see what we run into in the real world.

cjerdonek:
Thanks, @jungshadow.  Yes, I agree.  It's an iterative process.

Issue #40, opened 2015-03-07
Changelog missing from web documentation


cjerdonek:
The web documentation has a link to a "Changelog" in the left column under "Revision History," but there is no Changelog section.

jungshadow:
Hi, @cjerdonek. On a general note, it's great to have you digging in here. As for the issue at hand, since we've been on GitHub, we've retired the changelog for the issues and pull-request discussions as they tend to be substantially more meaningful. Apparently, we forgot to remove all of its remnants. We'll update the current documentation.

@nomadaisy Would you mind taking this one on, since you're currently working on some doc changes?

cjerdonek:
Hi @jungshadow, thanks, and you're welcome.  I do have a couple higher-level thoughts / questions about the organization of the various versions, but I'll hold off on mentioning those until I gain a little more familiarity with the project.

Regarding this issue, as an interested observer, I do think that a simple list of the changes from one version to the next would go a long way towards helping observers know how the versions differ from one another (e.g. attribute `foo` added, attribute `bar` removed or renamed, etc).  As it stands now, I couldn't find an easy way to uncover that information.  FWIW, on other projects I've been involved with, the changelog entries have links to the individual issue pages (when they exist), so that people can easily track down the context and discussion around the various changes.

jungshadow:
Feel free to reach out to us at info@votinginfoproject.org and set up a time to chat if you feel like it might be easier to discuss things over the phone :)

Hmm...I can understand the utility of the changelog. Personally, I'm not a fan of them because of the additional burden it puts on the team to re-digest and regurgitate–to use a rather unpleasant verb–a lot of the tracking that's already been done by the systems we use (e.g. GitHub, Basecamp [for general inter-team communication], et al). That said, I'd rather have the group decide if they want to keep one and it's conceivably easier to create one now that we mandate the use of milestones, issues, and pull-requests.

@pstenbjorn, @pkoms, @nomadaisy, @jdmgoogle: Being some of the more active folks on here, how do you feel about keeping a changelog?

pstenbjorn:
@jungshadow I am general agreement that for a schema a changelog is a bit of a burden, however, I think ensuring that pull requests and other commits have reasonably complete comments regarding the nature of the commits it would server a similar purpose.

cjerdonek:
[...]

PR #95, opened 2015-04-10
Stub out release notes for issue #40.

cjerdonek:
See issue #40.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,22 @@
+VIP Release Notes
+=================
+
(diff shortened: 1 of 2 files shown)


cjerdonek:
Merged per our conference call today.  Additional items can be added by posting a comment to issue #40.

jdmgoogle:
Change: Object identifiers are now of type xs:ID and references are of type xs:IDREF instead of xs:integer; this allows feed creators to perform cursory feed validation before publishing them.

PR #100, opened 2015-04-16
Added guide for potential contributors

### LINKED ISSUE #93: Add CONTRIBUTING file
To make sure we codify our current dev process, I'm going to add a CONTRIBUTING.md file to the vip5 branch.

jungshadow:
This should close #93 when merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,22 @@
+# Contributing to the VIP Specification
+Thanks for considering contributing to the Voting Information Project.
+
+While the following information shouldn't be considered hard-and-fast rules, they should be considered more than guidelines. As with anything in this repository, if you'd like to request a changes to this process, please fill out an issue and file it under the milestone "Up for Discussion."
+
+## Creating Issues
+
+* Like many open source projects, we strongly urge you to search through the existing issues before creating a new one.
+* Please include as many details as possible.
+    * Note that this specification's user base are the states and jurisdictions and its primary use case is to export structured election data out of the existing election management systems (e.g. voter registration systems, vote tabulation systems, et al).
+    * If you can include an example of how you think a new model should work, all the better.
+* All issues should be filed under the milestone "Up for Discussion" until the team moves it under a particular release or other related issue-management action.
+
+## Pull Requests
+
+1. Create a branch to work on a fix/feature (a fix/feature should have a companion bug/enhancement issue). Start the branch with either "feature/..." or "bug/...".
+    1. If you're not a member of the VIP Spec team, fork the repository and follow the same process.
+2. Once it's done and tested, create a pull request to move it into the current working branch.
+3. At that point, some discussion might happen and, when it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged into the current working branch by the developer who created the pull-request.
+4. Delete the feature/bug branch.
+
+At any one point in time ("feature/" and "bug/" temporary branches aside), there should only be a dev branch (called 'vip5' in the vip-specification's case, but it may change to simply 'dev' in the future), a stable branch (called 'master' in most cases), and, if necessary, a documentation branch (called 'gh-pages' if hosted on GitHub).


cjerdonek:
One thought: I was thinking this could go in the docs directory alongside the style guide.  (The "best practices" guide for feeds could also go in the docs directory.)  For more visibility, you could link to how to contribute in the README (GitHub supports internal relative links, e.g. `docs/contributing.md`).  By the way, I put "history" at the top level because I view it as project metadata at the same level as a readme or license file.

cjerdonek:
One more thought: you might want to link to the style guide in the contributing doc (e.g. in the bullets for creating a pull request).

jungshadow:
@cjerdonek Personally, I like the idea of `CONTRIBUTING.md` and `STYLEGUIDE.md` (not capitalized as of yet) being at the top-level of the repository as it details how to contribute and isn't documentation related to the specification (i.e. to me it falls into the project metadata category you mention).

cjerdonek:
@jungshadow I see your point about maintaining a distinction between external-facing docs and internal project docs, something I agree with.  I guess my point was more that when you start having more than one of something (e.g. "contributing" and "style guide"), I usually create a folder for it so as not to clutter the top level.  I was making a distinction between the history file and other internal project docs.

jungshadow:
I talked bit with @demcg about this on Friday and I believe we're of the same mind on this. On a related note, I'd like to create a separate repository to house the documentation, which will help with hosting/building this on [Read the Docs](https://readthedocs.org/). Any other thoughts on this? /cc @pstenbjorn @jdmgoogle @pkoms @nomadaisy

pstenbjorn:
@jungshadow I agree with using Read the Docs and having a separate Doc repo.

cjerdonek:
I'd strongly recommend keeping the docs in the same repo as the spec.  (This is also how most projects that I'm familiar with do it, as well as what Read The Docs says to do [in their documentation](https://read-the-docs.readthedocs.org/en/latest/getting_started.html).)

A couple advantages of doing it this way are that 1) the docs stay in synch with the version of the spec they are documenting (even when creating feature branches, etc), and 2) doc changes can be committed simultaneously with spec changes (which means you don't need to cross repos when adding to or viewing changes to the docs).

jungshadow:
@cjerdonek I didn't see a hard-and-fast recommendation in the Read the Docs documentation, but that sounds reasonable. Additionally, we can add to `CONTRIBUTING.md` that, with a spec change, the documentation must be kept up-to-date (once we have the documentation, that is).

So here's the plan:

1. Leave the `docs` folder for use for specification-related docs when we have them. I'm leaning toward [the Markdown option](https://read-the-docs.readthedocs.org/en/latest/getting_started.html#in-markdown).
2. Move STYLEGUIDE.md to the root (NB: All of the "metadocs" in the root are docs that I'd like any potential developer to read/access first. Additionally, if folks don't want to read through the `README`, they can have quick access to the metadocs of their particular focus/choice, using the directory listing as a makeshift menu of sorts).

Assuming that all makes sense, I'll merge today.

cjerdonek:
@jungshadow okay, thanks.

By the way, if you find that Markdown isn't flexible enough in terms of layout and styling options (e.g. in the way of providing what you already have [here](http://votinginfoproject.github.io/vip-specification/)), another possibility to look at for the longer term would be auto-generating the bulk of the HTML using templates.  For example, you could have an HTML template snippet for a single object type.  This would let you change the global structure and look-and-feel more easily by modifying things in one place.  This would take a bit more work to set up though, so would be a longer term project.

jungshadow:
I only merged `CONTRIBUTING.md`. To be radically clear, the other structural changes I talked about will be in another issue and PR.

jungshadow:
Interesting side note to this, after merging the `CONTRIBUTING.md` file, this alert now shows up on the issues/PR page. :+1: 

screen shot 2015-04-20 at 2 05 24 pm

PR #101, opened 2015-04-16
Sync the VIP spec with v41 of the 1622.2 standard.

jdmgoogle:
This syncs the draft 5.0 spec with v41 of the 1622.2 standard. There were some small changes around element and attribute names that I made in order to minimize the differences between the two. There will probably be one or two small changes in the coming days, but this should be the largest batch of changes.

@jungshadow @pstenbjorn and @cjerdonek Please review.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -267,7 +267,7 @@ reference elements for polling location
               <xs:element name="ElectoralDistrictID" type="xs:IDREF" />
               <xs:element name="Name" type="InternationalizedText" />
               <xs:element name="ExternalIdentifierID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="OfficeHolderID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
+              <xs:element name="OfficeHolderPersonID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="IsPartisan" type="xs:boolean" minOccurs="0" />
               <xs:element name="FilingDeadline" type="xs:dateTime" minOccurs="0" />
               <xs:element name="Term" type="Term" minOccurs="0" />
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -8,6 +8,7 @@ reference elements for polling location
   <xs:element name="VipObject">
     <xs:complexType>
       <xs:choice maxOccurs="unbounded">
+
         <!-- structural VIP elements-->
         <xs:element name="Source" minOccurs="1" maxOccurs="1">
           <xs:complexType>
@@ -189,7 +190,8 @@ reference elements for polling location
           <xs:complexType>
             <xs:sequence>
               <xs:element name="OrderedContest" type="OrderedContestType" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="InfoUrl" type="xs:anyURI" minOccurs="0" />
+              <xs:element name="PartyId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
+              <xs:element name="ImageUri" type="xs:anyURI" minOccurs="0" />
             </xs:sequence>
             <xs:attribute name="id" type="xs:ID" use="required" />
           </xs:complexType>
@@ -200,6 +202,7 @@ reference elements for polling location
             <xs:sequence>
               <xs:element name="PartyID" type="xs:IDREF" minOccurs="0" />
               <xs:element name="PersonID" type="xs:IDREF" minOccurs="0" />
+              <!-- SURVEY: Should external identifiers be referenced or included inline? -->
               <xs:element name="ExternalIdentifierID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="BallotName" type="InternationalizedText" />
               <xs:element name="FileDate" type="xs:dateTime" minOccurs="0" />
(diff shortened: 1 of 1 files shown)


vip_spec.xsd
```diff
@@ -189,7 +190,8 @@ reference elements for polling location
           <xs:complexType>
             <xs:sequence>
               <xs:element name="OrderedContest" type="OrderedContestType" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="InfoUrl" type="xs:anyURI" minOccurs="0" />
+              <xs:element name="PartyId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
```
cjerdonek: Casing: `PartyID`?

vip_spec.xsd
```diff
@@ -8,6 +8,7 @@ reference elements for polling location
   <xs:element name="VipObject">
     <xs:complexType>
       <xs:choice maxOccurs="unbounded">
+
```
cjerdonek: Are the random empty lines being added a style thing?  We should probably document that if so.

jdmgoogle:
Thanks. 1622 is using SomeId instead of SomeID. Do people have thoughts about changing to use that format?

vip_spec.xsd
jdmgoogle: I added the lines around the 'choice' tags because the end of the file was kind of a mess. The other blank lines I added were to create consistency with the other definitions which had blank lines between them. Although yes, we should add blank line use to the style guide.

vip_spec.xsd
```diff
@@ -244,11 +249,11 @@ reference elements for polling location
         <xs:element name="Contact">
           <xs:complexType>
             <xs:sequence>
-              <xs:element name="Url" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
+              <xs:element name="Uri" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="AddressLine" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="Email" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="Fax" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="HoursOpenID" type="xs:IDREF" minOccurs="0" />
+              <xs:element name="Schedule" type="xs:IDREF" minOccurs="0" />
```
cjerdonek: It seems like this should be `HoursOpenID` to match our style guide?  See [here](https://github.com/votinginfoproject/vip-specification/blob/vip5/docs/style_guide.md#ids).

vip_spec.xsd
```diff
@@ -260,13 +265,14 @@ reference elements for polling location
...
             <xs:sequence>
               <xs:element name="ElectoralDistrictID" type="xs:IDREF" />
-              <xs:element name="ContactID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="Name" type="InternationalizedText" />
               <xs:element name="ExternalIdentifierID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="Partisan" type="xs:boolean" minOccurs="0" />
-              <xs:element name="TermEndDate" type="xs:date" minOccurs="0" />
-              <xs:element name="TermStartDate" type="xs:date" minOccurs="0" />
-              <xs:element name="TermType" type="TermType" />
+              <xs:element name="OfficeHolder" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
```
cjerdonek: Again, our style is to end in `[TypeName]ID` for ID references.  (By the way, I do think the guideline helps because it says what type the ID is a reference to.)

cjerdonek:
> Do people have thoughts about changing to use that format?

I'd be okay with either.

pstenbjorn:
I am happy with either, visually I prefer `[TypeName]ID` but `[TypeName]Id` seems to be what 1622.2 is standardizing around.

jdmgoogle:
Okay, I've made the recommended updates. Once this PR goes in I'm going to do another round of changes to the XSD and the style guide to change things like ReferenceToID to ReferenceToId just to be consistent with 1622.

jdmgoogle:
So 1622.2 added the concept of the current office holder to the Office object (there was some misunderstanding in what I was asking for, but I actually like it). The problem is that if we just use PersonID it's ambiguous as to which person this means. Suggestions?

cjerdonek:
I would prefix it then, perhaps "OfficeHolderPersonID".  The style guide anticipated this by saying that the  attribute name should end with the type name followed by "ID".

vip_spec.xsd
```diff
@@ -200,6 +202,7 @@ reference elements for polling location
             <xs:sequence>
               <xs:element name="PartyID" type="xs:IDREF" minOccurs="0" />
               <xs:element name="PersonID" type="xs:IDREF" minOccurs="0" />
+              <!-- SURVEY: Should external identifiers be referenced or included inline? -->
```
cjerdonek: Personally, I think inline makes more sense.  I also think we should define an "ExternalIdentifiers" (or "ExternalIdentifierCollection") type since everywhere we want to allow external identifiers, it will always be a collection.

jdmgoogle: Okay. I think that should be a different PR, but I just wanted to get everyone's opinion here.

jdmgoogle:
Once @pstenbjorn and @jungshadow sign off on this, I'm going to merge this PR.

vip_spec.xsd
cjerdonek: Cool, do you want to create the issue?

pstenbjorn:
@jdmgoogle I am fine with the changes and I agree that the ExternalIdentifiersCollection concept should be handled in a separate PR.

jungshadow:
@jdmgoogle I'd like to look at this, but I also don't want to unreasonably hold this up (i.e. it may still take me a few hours to get to as I'm not in the office). If you have @cjerdonek's and @pstenbjorn's approval on this, probably safe to say it's solid and any niggling bits I might bring up won't be show-stoppers and potentially fixed in subsequent updates.

jdmgoogle:
Okay. Merging and moving on to the next issue.

pstenbjorn:
@jdmgoogle would you like me to take a first stab at Venetian Blind conversion next?

jdmgoogle:
@pstenbjorn Let me make one or two other small changes and then yes, feel free to take a stab at the VB conversion.

PR #103, opened 2015-04-17
Update style guide and XSD to be ReferencedTypeId instead of ReferencedTypeID

jdmgoogle:
This is a small change for 1622.2 compatibility.


cjerdonek:
LGTM

Issue #63, opened 2015-03-26
Various voter services are handled by different local offices at the locality level


jungshadow:
Even after many years in elections, there's always something new to discover :) Just had a conversation with Alabama. Some voter services are run by different entities and sometimes in separate buildings. For example:

* The Probate Judge is the chief election official in the county who deals with the polling places, ballots, and running the election;
* The Board of Registrars, typically 3 or 4 member board, handles voter registration
* The Absentee Election Manager handles all absentee ballots

Currently, VIP can surface information on an the election official and an "overseas vote contact" (the latter probably inaccurately named), but they all link to a single physical address. Even with this model, we still couldn't present a contact for registration information (NB: many of these entities lack websites or online services).

Clearly, it would be ideal if we could route people to one contact, but, in this case, this is just going to add administrative overhead on the official who has to redirect voters and this will probably frustrate the voter. I'd welcome some thoughts on how to approach this.

cjerdonek:
@jungshadow Based on your description, to address cases where the entities are completely different, it seems like it would make sense to have an object type corresponding to "entity responsible for some class of voter services," and then have the spec allow including more than one such object per jurisdiction.  The object type could have a free text field describing what they are responsible for (or perhaps an enum, but there you run into having to know the possible descriptions / combinations in advance, which is probably impossible).  Something like that seems potentially straightforward.

The other use case I can think of is if a single Department or organization is responsible for all of the voter services, but they might want to route someone to a specific part of that organization (e.g. a specific person or phone number, etc).  For example, the SF Department of Elections has a [list here](http://www.sfgov2.org/index.aspx?page=3335) of several phone numbers, e.g.

> For questions about voter registration and vote-by-mail voting, 
(415) 554-4411
For Voter Information Pamphlets, 
(415) 554-4375
For information about becoming an Elections Officer (Pollworker), 
(415) 554-4395
For information about candidates or campaign disclosure reports, 
(415) 554-4375
For election results on Election Night, 
(415) 554-4375 
or this website
For a free workshop regarding current election information, 
(415) 554-6184
For information about polling places and ADA issues, 
(415) 554-4551

To address this case, it seems like a simple list of entities might not completely suffice (because you would want the information to reflect that these are all parts of a single entity, so there might be some additional relationships to capture).

djbridges:
This disaggregated responsibility (as well as inconsistency in
responsibility assignments even intra-state) is true in many instances -
[...]

PR #104, opened 2015-04-17
Reuse some 1622.2 objects in VIP

jdmgoogle:
This PR needs a bit of love and attention, but the basic plan is to reuse 1622.2 objects within VIP where possible, and clean up some of the references, variable names, and links.

This also implements a simple solution to #63 by making ElectionAdministrationId a repeated field at the state and local level. Combined with the VoterServices field, this may be "good enough" for our purposes.

Cc: @jktomer @pstenbjorn @jungshadow @cjerdonek @pkoms @djbridges

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -31,12 +31,12 @@ reference elements for polling location
               <xs:element name="Date" type="xs:date" />
               <xs:element name="ElectionType" type="InternationalizedText" minOccurs="0" />
               <xs:element name="StateId" type="xs:IDREF" />
-              <xs:element name="Statewide" type="xs:boolean" minOccurs="0" />
+              <xs:element name="IsStatewide" type="xs:boolean" minOccurs="0" />
               <xs:element name="RegistrationInfo" type="InternationalizedText" minOccurs="0" />
               <xs:element name="AbsenteeBallotInfo" type="InternationalizedText" minOccurs="0" />
               <xs:element name="ResultsUrl" type="xs:anyURI" minOccurs="0" />
               <xs:element name="PollingHours" type="InternationalizedText" minOccurs="0" />
-              <xs:element name="ElectionDayRegistration" type="xs:boolean" minOccurs="0" />
+              <xs:element name="HasElectionDayRegistration" type="xs:boolean" minOccurs="0" />
               <xs:element name="RegistrationDeadline" type="xs:date" minOccurs="0" />
               <xs:element name="AbsenteeRequestDeadline" type="xs:date" minOccurs="0" />
             </xs:all>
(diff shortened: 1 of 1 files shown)


cjerdonek:
Regarding the #63 aspect of this PR, it seems like each locality should only have one object that contains all of the standard URLs.  Otherwise, you run into issues like which ElectionAdministration object to put the URLs in.  And why spread the URLs over multiple ElectionAdministration objects?

It seems like #63 could be handled pretty simply just by having a "VoterServices" type.  The type would have an enum or free-form description saying what the service is, along with a ContactID (which already contains address, phone, hours, etc).  Then a locality would simply list one "VoterServices" object for each service it offered.  The URLs could stay in the one ElectionAdministration object.

jdmgoogle:
I did it this way in case the different administrative duties. For example, in @jungshadow's example in #63, one department handles polling places, ballots, and running the election (the WhereDoIVoteUrl, WhatIsOnMyBallotUrl, and ElectionsUrl); another handles voter registration (RegistrationUrl and AmIRegisteredUrl); and another handles absentee ballots (AbsenteeUrl).

I acknowledge that this opens up problems if there are multiple administrative departments that have the same URL field filled out (e.g., two ElectionAdministration objects, each with unique WhereDoIVoteUrl elements) but that's something which could be caught during feed validation.

cjerdonek:
I'm not sure it's important to associate an URL with a particular Department, but if so, that could still be done with what I'm suggesting.  What I'm primarily advocating for is a list of light-weight "VoterService" objects in each locality containing an enum / description and a reference to a Contact object of some sort. 

Here is an example I mentioned in the original thread (the [first comment](https://github.com/votinginfoproject/vip-specification/issues/63#issuecomment-86687460) in the thread, actually) where the proposal in the current PR seems to break down.  The example is where there may be several services within a Department, but each differs in only a small bit of info (e.g. phone number or room number).  See here, for example: http://www.sfgov2.org/index.aspx?page=3335

With the PR, you'd have to add a heavy-weight ElectionAdministration object for every one of these services to differentiate the contact info, but they would all correspond to the same Department.  In addition, that would create the problem of which one of these redundant ElectionAdministration objects would you put the URLs in.

In my proposal, you would just create a new Contact object for each service that required new contact info, without duplicating the Department info.

jdmgoogle:
What if we added an "AdditionalNotes" field to the Contact object and put the voter services there? Then we could have the URLs in the main administration object and repeated Contact objects with different strings in the AdditionalNotes field?

cjerdonek:
From a modeling perspective, I think that is an improvement, though I'm not sure to what extent the services need to accessed in a structured way (e.g. via an enum).  You and @jungshadow would know more about that.  For example, it seems like it would be messy if you were trying to parse individual service names out of a free-form field.

I get the sense that you're looking for a solution that doesn't involve introducing new type definitions.  Is that the case?

cjerdonek:
FYI, I noticed that the Contact type itself has a "Uri" element.  So it seems like many of the "service URLs" could even be accommodated by voter service objects of the form "(service enum/description, contact_id)."  Not necessarily advocating that, but just thought I'd mention the possibility.

cjerdonek:
I took a own crack at addressing issue #63 here: #105.

I added the ability to add multiple "Departments" to the ElectionAdministration container object.  Moreover, each Department can have multiple voter services, with customizable contact info for each.

jdmgoogle:
@cjerdonek I removed the election administration aspect of this, so this PR should be orthogonal to #105 now. Please take another look.

Cc @pstenbjorn @jktomer @jungshadow

cjerdonek:
@jdmgoogle thanks.  Looks okay to me :+1: 

Btw, the "Is / Has" language might be good for the style guide.

jungshadow:
A bit late on this, but looks good to me. A bit of a pedantic note, but, without writing this in stone, can we at least try to get sign-off on structural changes from at least two people before merging? Also, does this happen to close #63 (or at least convert it to an accepted v5 issue) or other related issues?

Just want to make sure we're setting a good example and following our own guidelines :)

Issue #112, opened 2015-04-20
Remove duplication between SimpleAddressType and Contact types


cjerdonek:
It looks like there is duplication between the SimpleAddressType and Contact types.

cjerdonek:
It looks like SimpleAddressType is only used by PollingLocation, which also has HoursOpenId.  So one possibility would be to change PollingLocation to use Contact instead.

In addition, another possibility would be to have Contact adopt SimpleAddressType's more structured form of address, instead of its current unstructured form:

```xml
<xs:element name="AddressLine" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
```

jungshadow:
@cjerdonek The latter option seems like the most flexible and the "best of both worlds."

cjerdonek:
Okay, I will make a PR.

jungshadow:
@cjerdonek Great!

cjerdonek:
Created PR #114.

jdmgoogle:
TIMEOUT! :)

I understand the desire to have a single address type in the spec, but unfortunately (for now) we need to have Contact use multiple lines (for 1622.2 compatibility) and VIP use the SimpleAddressType.

The TL;DR version is that 1622.2 will likely the basis of an international election results reporting standard, one for countries for which the "city, state, zip" framework doesn't work, whereas VIP continues to need to support street segments for its core mission of polling place lookups.

My counter-proposal is to open a post-5.0 issue to locate and use one address standard which fits both of these use cases. So please drop #114. and I'll file a new issue as a placeholder for a long-term address standard.

cjerdonek:
> My counter-proposal is to open a post-5.0 issue to locate and use one address standard which fits both of these use cases.

FYI, It's a bit aggressive to close this issue and reopen one that is basically the same.  I think the more appropriate thing to do would have been to continue the discussion here and change the milestone fro 5.0 to post 5.0.  Why was this issue closed?

jdmgoogle:
I suppose I could have kept this one open and renamed it. Primarily I wanted something different to reflect the different and larger scope of the issue. Removing duplication between the two objects is an action that will flow out of the address specification evaluation process, but the actual activity that needs to happen before that is to identify and evaluate different addressing standards.

cjerdonek:
Okay, I can see how the scope of that issue is more general.  I do think, though, that as a matter of courtesy you should discuss before immediately closing.

[...]

PR #114, opened 2015-04-20
Address issue #112: make PollingLocation more like Contact

cjerdonek:
This addresses issue #112.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -166,8 +166,7 @@ reference elements for polling location
                    in future versions of VIP.
               -->
               <xs:element name="Hours" type="InternationalizedText" minOccurs="0" />
-              <!-- Question: should unbounded be removed as in Contact? -->
-              <xs:element name="HoursOpenId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
+              <xs:element name="HoursOpenId" type="xs:IDREF" minOccurs="0" />
               <xs:element name="PhotoUrl" type="xs:anyURI" minOccurs="0" maxOccurs="1" />
               <xs:element name="IsEarlyVoting" type="xs:boolean" minOccurs="0" maxOccurs="1" />
               <xs:element name="IsDropBox" type="xs:boolean" minOccurs="0" maxOccurs="1" />
@@ -275,7 +274,7 @@ reference elements for polling location
               -->
               <xs:element name="Hours" type="InternationalizedText" minOccurs="0" />
               <xs:element name="HoursOpenId" type="xs:IDREF" minOccurs="0" />
-              <!-- Question: is this a person or place name? -->
+              <!-- This can be a person or place name. -->
               <xs:element name="Name" type="xs:string" minOccurs="0" />
               <xs:element name="Phone" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="Fax" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -159,12 +159,18 @@ reference elements for polling location
         <xs:element name="PollingLocation">
           <xs:complexType>
             <xs:sequence>
-              <xs:element name="Address" type="SimpleAddress" />
+              <!-- It is intentional that an address is required. -->
+              <xs:element name="AddressLine" type="xs:string" maxOccurs="unbounded" />
               <xs:element name="Directions" type="InternationalizedText" minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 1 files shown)


jdmgoogle:
Cut-and-pasting my comment from #112:

I understand the desire to have a single address type in the spec, but unfortunately (for now) we need to have Contact use multiple lines (for 1622.2 compatibility) and VIP use the SimpleAddressType.

The TL;DR version is that 1622.2 will likely the basis of an international election results reporting standard, one for countries for which the "city, state, zip" framework doesn't work, whereas VIP continues to need to support street segments for its core mission of polling place lookups.

My counter-proposal is to open a post-5.0 issue to locate and use one address standard which fits both of these use cases. So please drop #114. and I'll file a new issue as a placeholder for a long-term address standard.

pstenbjorn:
@jdmgoogle I second your proposal.

cjerdonek:
> I understand the desire to have a single address type in the spec, but unfortunately (for now) we need to have Contact use multiple lines (for 1622.2 compatibility)

It seems like 1622.2 has different requirements if it is international.  What's wrong with being US-centric on the VIP side since we can, and then doing a simple conversion to 1622.2's format if necessary?  This would be possible because the VIP feeds would be _more_ structured rather than less.  I'm assuming the VIP feeds are limited to the USA?

The conversion from VIP to 1622.2, for example, could be to concatenate the City, State, Zip info into a single line for 1622.2's generic "AddressLines" unbounded element.
 
Also, why wait until after 5.0?  By waiting, departments and office-holders, etc. won't have city, state, zip broken out which will make the feed harder to parse.

jdmgoogle:
The all-lines-are-generic approach will not work for street segments since street segments are US-centric and require a city/state/zip breakdown to work effectively.

The all-lines-have-detail approach will not work for Contact information since that will add complexity to election departments which are attempting to support both VIP and 1622, and the Contact elements in 1622 are not broken down. This means that they'll need to support two different methods of exporting contact information, which will add even more complexity to an already-complex process; @pstenbjorn can provide more detail since he's working on this exact issue for several states.

cjerdonek:
> This means that they'll need to support two different methods of exporting contact information

With what I'm suggesting, they would only need to provide the data in one format: with the City, State, Zip broken out.  Since more structure is being preserved in the export, the other format can be generated by conversion automatically.

jdmgoogle:
This is assuming that all the data about all contact information -- including data they get from campaigns, etc -- is already provided and stored natively in city, state, zip format.

jdmgoogle:
Note to all involved: the discussion on the larger issue has been moved back to issue #112.

cjerdonek:
I updated the PR based on the issue #112 discussion and retitled the PR to reflect the change.

vip_spec.xsd
```diff
@@ -159,12 +159,18 @@ reference elements for polling location
...
             <xs:sequence>
-              <xs:element name="Address" type="SimpleAddressType" />
+              <!-- It is intentional that an address is required. -->
+              <xs:element name="AddressLine" type="xs:string" maxOccurs="unbounded" />
               <xs:element name="Directions" type="InternationalizedText" minOccurs="0" maxOccurs="1" />
+              <!-- Note: The "Hours" element is being deprecated and will be removed
+                   in future versions of VIP.
+              -->
+              <xs:element name="Hours" type="InternationalizedText" minOccurs="0" />
+              <!-- Question: should unbounded be removed as in Contact? -->
```
jdmgoogle: Yes.

vip_spec.xsd
```diff
@@ -263,17 +269,18 @@ reference elements for polling location
...
-              <xs:element name="Uri" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="AddressLine" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="Email" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
-              <xs:element name="Fax" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
               <!-- Note: The "Hours" element is being deprecated and will be removed
                    in future versions of VIP.
               -->
               <xs:element name="Hours" type="InternationalizedText" minOccurs="0" />
               <xs:element name="HoursOpenId" type="xs:IDREF" minOccurs="0" />
+              <!-- Question: is this a person or place name? -->
```
jdmgoogle: It can be either.

jdmgoogle:
Looks Good To Me.

Comments, @jungshadow @pstenbjorn @jen-tolentino?

pstenbjorn:
@jdmgoogle LGTM

cjerdonek:
Thanks a lot for the review, @jdmgoogle and @pstenbjorn.

Issue #106, opened 2015-04-19
Style guide should state the "Other" pattern used with enums


cjerdonek:
The style guide should describe the following pattern used throughout the spec of adding an "Other" element with type `xs:string` whenever we want to allow values not included in an enumeration:

```xml
<xs:element name="VoteVariationType" type="VoteVariation" minOccurs="0" />
<xs:element name="OtherVoteVariationType" type="xs:string" minOccurs="0" />
```

cjerdonek:
Created PR #115.

jdmgoogle:
Looks Good To Me

cjerdonek:
Thanks. @pstenbjorn how does this look to you? @jungshadow requested sign off from two people.

cjerdonek:
Closed by PR #115.

PR #115, opened 2015-04-20
Address issue #106: document style for "other" and enums

cjerdonek:
Address issue #106.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -70,18 +70,21 @@ For example--
 ```
 
 
-### Enumerations and "Other"
+### Enumeration Ordering and "Other"
 
-To support "other" values in an enumeration, add the value "other"
-to the enumeration type definition.  For example--
+Enumeration values should be arranged alphabetically, except for "other",
+which should appear last.
+
+To support "other" values in an enumeration, include "other" as a value.
+For example--
 
 ```xml
 <xs:simpleType name="Color">
   <xs:restriction base="xs:string">
     <xs:enumeration value="blue" />
     <xs:enumeration value="green" />
-    <xs:enumeration value="other" />
     <xs:enumeration value="red" />
+    <xs:enumeration value="other" />
   </xs:restriction>
 </xs:simpleType>
 ```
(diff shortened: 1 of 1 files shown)


cjerdonek:
@pstenbjorn, @jungshadow approve?

STYLEGUIDE.md
```diff
@@ -69,6 +70,32 @@ For example--
...
+
+To support "other" values in an enumeration, add the value "other"
+to the enumeration type definition.  For example--
+
+```xml
+<xs:simpleType name="Color">
+  <xs:restriction base="xs:string">
+    <xs:enumeration value="blue" />
+    <xs:enumeration value="green" />
+    <xs:enumeration value="other" />
```
jungshadow: For general usability, I might also add that "other" should be the last option. While I understand having the options in alphabetical order, I think the "other" might be lost on a longer list. Maybe alphabetical-with-an-other?

cjerdonek: I like that idea.  Thanks.

STYLEGUIDE.md
```diff
@@ -51,7 +52,7 @@ Markdown, etc).
     ````
 
 
-### Attributes
+### Attribute Ordering
```
jungshadow: Why the name change?

jungshadow:
Name question aside, looks good.

STYLEGUIDE.md
cjerdonek: I did that because attributes are also discussed elsewhere in the document, and this section is only about ordering.

Actually, there is also a bit of a confusion in that "attribute" is used in two senses in the style guide: for attribute definitions (i.e. `xs:attribute`) and also attributes of tags within the XSD itself.  This section is about the latter, so perhaps an even better name can be used (to distinguish from, say, how `<xs:attribute>` elements should be ordered within a section, which is another topic that has crossed my mind).

cjerdonek:
Updated to address @jungshadow's comments.  I will commit this tomorrow unless I hear any objection.  @jdmgoogle already gave approval to the initial version [here](https://github.com/votinginfoproject/vip-specification/issues/106#issuecomment-94746474).

jdmgoogle:
Just to be clear: LGTM. Approved.

cjerdonek:
Thanks

Issue #30, opened 2015-03-04
Limit the number of types for electoral_district


jungshadow:
The element `electoral_district.type` is currently a string, which allows any value. This could be an enumerated list of types (e.g. state senate, congressional, et al...). The data in state systems isn't standardized and this could be difficult on VIP's end to structure, but it's easy to see the benefits of structuring the information. This issue is related to #29.

pstenbjorn:
If the use of ocdids is pursued, the enumerations for electoral_district.type might be inherited from the ocdid keys.  E.g. in the ocdid
division/country:us/state:ma/county:norfolk/school_district:cohasset
There are four district types identified, country, state, county, and school_district.  

As @jungshadow mentioned, each state may have its own localization of a type and we have discussed adding electoral_district.localization to accommodate this.

jdmgoogle:
In my opinion there are two ways to do this. The first way is to use a subset of the enums in the ReportingUnit type in 1622.2; this list includes:

City
Congressional
Council
County
Judicial
Local
Locality
Municipality
National
Other
School
Special
State
StateHouse
StateSenate
Township
Utility
Ward
Water

Option #2 is to use the set of types used in OCD ID keys, per @pstenbjorn's suggestion. There's obviously a lot of overlap between the two, but I'd like to poll people here for their recommendation.

Cc @jungshadow @cjerdonek @nomadaisy @pkoms @demcg @jktomer @jen-tolentino

cjerdonek:
For convenience, here is the current definition:

```xml
<xs:element name="ElectoralDistrict">
  <xs:complexType>
    <xs:sequence>
      <xs:element name="Name" type="xs:string" />
      <xs:element name="Type" type="xs:string" minOccurs="0" />
      <xs:element name="Number" type="xs:integer" minOccurs="0" />
      <xs:element name="ExternalIdentifierID" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
    </xs:sequence>
    <xs:attribute name="id" type="xs:ID" use="required" />
  </xs:complexType>
</xs:element>
```

Is it the intent that there be a way to group districts together of a common type (e.g. all supervisorial districts)?  And is this what the "Type" name is for (via string matching), or is it just for display?

pstenbjorn:
For reference purposes these are the implicit types in the OcdId model.  @cjerdonek I know that currently their are strict enumerations for election type used for presentation logic on Google's side.

court_of_appeals
territory
country
district
state
[...]

PR #119, opened 2015-04-21
Add a DistrictType enumeration and use that in ElectoralDistrict elements

jdmgoogle:
This is a fix for #30.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -558,7 +558,6 @@ reference elements for polling location
       <xs:enumeration value="locality" />
       <xs:enumeration value="municipality" />
       <xs:enumeration value="national" />
-      <xs:enumeration value="other" />
       <xs:enumeration value="school" />
       <xs:enumeration value="special" />
       <xs:enumeration value="state" />
@@ -568,6 +567,7 @@ reference elements for polling location
       <xs:enumeration value="utility" />
       <xs:enumeration value="ward" />
       <xs:enumeration value="water" />
+      <xs:enumeration value="other" />
     </xs:restriction>
   </xs:simpleType>
 
@@ -594,9 +594,9 @@ reference elements for polling location
     <xs:restriction base="xs:string">
       <xs:enumeration value="ballot-measure" />
       <xs:enumeration value="initiative" />
-      <xs:enumeration value="other" />
       <xs:enumeration value="referendum" />
       <xs:enumeration value="retention" />
+      <xs:enumeration value="other" />
     </xs:restriction>
   </xs:simpleType>
 
@@ -636,23 +636,23 @@ reference elements for polling location
       <xs:enumeration value="cumulative" />
       <xs:enumeration value="majority" />
       <xs:enumeration value="n-of-m" />
-      <xs:enumeration value="other" />
       <xs:enumeration value="plurality" />
       <xs:enumeration value="proportional" />
       <xs:enumeration value="range" />
       <xs:enumeration value="rcv" />
       <xs:enumeration value="rcv-borda" />
       <xs:enumeration value="super-majority" />
+      <xs:enumeration value="other" />
     </xs:restriction>
   </xs:simpleType>
 
   <xs:simpleType name="VoterServiceType">
     <xs:restriction base="xs:string">
       <xs:enumeration value="absentee-ballots" />
-      <xs:enumeration value="other" />
       <xs:enumeration value="overseas-voting" />
       <xs:enumeration value="polling-places" />
       <xs:enumeration value="voter-registration" />
+      <xs:enumeration value="other" />
     </xs:restriction>
   </xs:simpleType>
   <!-- End of enumerated values -->
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -123,7 +123,7 @@ reference elements for polling location
               <!-- A locality may have more than one department with each handling different services. -->
               <xs:element name="Department" maxOccurs="unbounded">
                 <xs:complexType>
-                  <xs:all>
+                  <xs:sequence>
                     <xs:element name="ContactInformationId" type="xs:IDREF" minOccurs="0" />
                     <xs:element name="ElectionOfficialPersonId" type="xs:IDREF" minOccurs="0" />
                     <xs:element name="VoterService" minOccurs="0" maxOccurs="unbounded">
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -174,7 +174,8 @@ reference elements for polling location
           <xs:complexType>
             <xs:sequence>
               <xs:element name="Name" type="xs:string" />
-              <xs:element name="Type" type="xs:string" minOccurs="0" />
+              <xs:element name="Type" type="DistrictType" />
+              <xs:element name="OtherType" type="xs:string" minOccurs="0" />
               <xs:element name="Number" type="xs:integer" minOccurs="0" />
               <xs:element name="ExternalIdentifiers" type="ExternalIdentifiers" minOccurs="0" />
             </xs:sequence>
@@ -576,6 +577,30 @@ reference elements for polling location
       <xs:enumeration value="unofficial-partial" />
       <xs:enumeration value="unofficial-complete" />
       <xs:enumeration value="certified" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="DistrictType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="city" />
+      <xs:enumeration value="congressional" />
+      <xs:enumeration value="council" />
+      <xs:enumeration value="county" />
+      <xs:enumeration value="judicial" />
+      <xs:enumeration value="local" />
+      <xs:enumeration value="locality" />
+      <xs:enumeration value="municipality" />
+      <xs:enumeration value="national" />
+      <xs:enumeration value="other" />
+      <xs:enumeration value="school" />
+      <xs:enumeration value="special" />
+      <xs:enumeration value="state" />
+      <xs:enumeration value="statehouse" />
+      <xs:enumeration value="statesenate" />
+      <xs:enumeration value="township" />
+      <xs:enumeration value="utility" />
+      <xs:enumeration value="ward" />
+      <xs:enumeration value="water" />
     </xs:restriction>
   </xs:simpleType>
 


jdmgoogle:
Pinging @jktomer @jungshadow @pstenbjorn @cjerdonek and others for review.

pstenbjorn:
@jdmgoogle looks good to me.

vip_spec.xsd
```diff
@@ -579,6 +580,30 @@ reference elements for polling location
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="DistrictType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="city" />
+      <xs:enumeration value="congressional" />
+      <xs:enumeration value="council" />
```
jktomer: Sorry for missing this in my earlier review of the 1622 list. "Council" is insufficient to disambiguate between city council and county council elections, which quite commonly occur together. That means that in an election where there are both city- and county-council contests, and both VIP and unofficial contest data, we would be unable to serve the VIP data even if it includes unique contests.

I'm not aware of any other commonly-coöccuring pair of district types that would present this problem.

cjerdonek: > we would be unable to serve the VIP data even if it includes unique contests.

Hmm, I thought I asked basically this same question before [here](https://github.com/votinginfoproject/vip-specification/issues/30#issuecomment-93998098), and @jdmgoogle replied that it was okay.  He replied--

> No, there's no problem if a precinct links to more than one electoral district of the same type.

cjerdonek:
Since we already decided to use 1622.2's enumeration, should we be reviewing the list itself, or is that set in stone?

vip_spec.xsd
```diff
@@ -579,6 +580,30 @@ reference elements for polling location
...
+      <xs:enumeration value="county" />
+      <xs:enumeration value="judicial" />
+      <xs:enumeration value="local" />
+      <xs:enumeration value="locality" />
+      <xs:enumeration value="municipality" />
+      <xs:enumeration value="national" />
+      <xs:enumeration value="other" />
```
cjerdonek: This should go at the end per the style guide.

vip_spec.xsd
jdmgoogle: I think we're talking about slightly different things. I was talking about what would happen if there was a precinct that linked two Electoral Districts *actually* of the same type; that's not a problem. What @jktomer is talking about is a precinct linking to two Electoral Districts which are *not actually* of the same type, combined with multiple sources of ballot information, combined with the current internal logic of our pipeline for de-duplicating from those two sources.

cjerdonek: Okay, the impression I got from that reply was that it wasn't important that the enum values be unique as a differentiator.  The issue @jktomer raised is one of the reasons why I was [suggesting](https://github.com/votinginfoproject/vip-specification/issues/30#issuecomment-93985805) the idea to have a district type object (as opposed to enum).  It would be guaranteed to be unique, in addition to capturing more of the relationships.

cjerdonek: > but for which we don't have sufficiently unique enum values

Also, how can you ever know in advance that the enum values are sufficiently unique?  For example, I'm guessing there is probably more than one kind of "judicial" district in some parts of the country.  In California, the state has six Courts of Appeal districts, but maybe some parts of the country also have more localized judicial districts.  The SF Bay Area also has BART districts (a type of transportation district), which I guess would be classified as an under in this scheme.  California also has four "Board of Equalization" districts, which I guess would be another "other."

jktomer: Actually, I might have been a bit too paranoid here.

Our actual logic looks like this:
* Group contests in each dataset by electoral district type (*only* -- other identifiers, like district name, are considered too unreliable across differing data sources; the only thing that's likely to change my mind on that front is if we start getting OCD IDs from many or most VIP participants).

* For each electoral district type, pick the "best" dataset for contests in districts of that type based on request parameters and some heuristics, and serve all the contests from that dataset for that district.

So the cost of having city- and county-council districts share an electoral district type is that we will only serve contests from a single dataset for such a type. In a typical situation (where we have contests from exactly two datasets -- VIP and an unofficial source), this will make little difference in practice: our unofficial dataset always has both city- and county-council contests, and we generally would prefer it for both unless the user has specifically requested only official data, in which case there's no race.

However, as it becomes more common for multiple overlapping VIP datasets to provide information about a single address (for example, some states have a single dataset with statewide contests, and then a dataset for each county), finer-grained district identifiers become potentially more important. (Note that I'd still far prefer relying on OCD IDs than a new custom district identification scheme.) For now, though, it still seems highly unlikely (and suspect) for one VIP dataset  to provide county-council contests and another to provide city-council districts. So I withdraw my objection, unless you can convince the 1622.2 committee to make a last-minute editorial change here.

> Also, how can you ever know in advance that the enum values are sufficiently unique? For example, I'm guessing there is probably more than one kind of "judicial" district in some parts of the country.

You're right, but in practice, we had a quite limited set of district types before and never encountered collisions, at least not for judicial districts. That's a far cry from ultimate safety, but absent emprical evidence of collisions I'm inclined to leave this alone and instead push for better OCD ID adoption.

cjerdonek: Hmm, this all seems a bit brittle to me.  It also seems unclear how a dataset is supposed to choose the enumeration value (i.e. so that different datasets would be using the same value).  For example, what is the difference between city, local, locality, and municipality (and county in the case of a City and County like San Francisco)?  Or utility, water, other?  Or council and ward?  And what about when there are multiple "other" values (e.g. BART and the Board of Equalization examples I mentioned above)?

FWIW, I did find what seem to be at least three types of judicial districts in California, each with its own numbering sequence (presumably all applicable in retention elections). The California courts of appeal has six districts within the state. Los Angeles County Superior Court of California has thirteen judicial districts.  And the Fourth District of the California Court of Appeal has three districts within that.

For "school," note that there are overlapping districts here as well (e.g. Community College districts and the school districts for lower levels like elementary and high school).

Also, just to be clear, my original suggestion wasn't to invent a new identification scheme, but to introduce a parent object that would serve to group the districts within a numbering sequence.  It is agnostic to the particular identification scheme: it doesn't require one but could still be used in conjunction with one (and in conjunction with 1622.2).

jdmgoogle: There are two basic responses to these comments, and I'll try and address each of your concerns.

The first issue is "should we use 1622.2 enumerations?" I believe we should for the sake of compatibility and the increased likelihood that we'll get more data as a result of this compatibility.

The second issue is "does 1622.2 contain the right set of enumerations?" And here I agree that this list could use work.

> For example, what is the difference between city, local, locality, and municipality (and county in the case of a City and County like San Francisco)?

WRT city versus local or locality or municipality, I can see the former used when there's an actual city, and the latter used when it's a sub-city district or a township. I agree, however, that these enumerations may be overly specific and are probably due to someone in 1622.2 saying that "well, we don't have *local* elections, we have *locality* elections, so that needs to be separate" rather than recognizing that these are "close-enough" labels.

> FWIW, I did find what seem to be at least three types of judicial districts in California, each with its own numbering sequence (presumably all applicable in retention elections).

I think this echoes our earlier conversation about offices versus the governmental body to which those offices are related. The enumeration should *only* establish the *type* of the office, not attempt to define the overall structure of government. This distinction should go in the eventual "Best Practices" document, as this would be a bad object:
```
  <ElectoralDistrict>
    <Name>Fourth District</Name>
    <Type>judicial</Type>
  </ElectoralDistrict>
```
whereas this would be better
```
  <ElectoralDistrict>
    <Name>California Court of Appeals, Fourth District</Name>
    <Type>judicial</Type>
  </ElectoralDistrict>
```

> my original suggestion wasn't to invent a new identification scheme, but to introduce a parent object that would serve to group the districts within a numbering sequence.

My concern here is that this simply moves the disambiguation problem up a level and doesn't necessarily help us with the underlying "feed publisher includes insufficient data" problem. Someone could still create an ElectoralDistrict named "Second District" of type "judicial" and have its parent be "Fourth District" (also of type "Judicial") and we're no closer to actually figuring out how to disambiguate and resolve these problems.

I propose we

1) Include the current set of 1622.2 enumerations as-is; and
2) Create a concrete list of suggestions for the 1622.2 schema (e.g., "either document specific examples and provide guidelines on how to select city vs locality vs municipality, or drop possibly-duplicate values") and then track and resulting changes.

vip_spec.xsd
jdmgoogle: Thanks. Will update in next commit.

vip_spec.xsd
jktomer: > Hmm, this all seems a bit brittle to me. It also seems unclear how a dataset is supposed to choose the enumeration value (i.e. so that different datasets would be using the same value). For example, what is the difference between city, local, locality, and municipality (and county in the case of a City and County like San Francisco)?

Indeed, I expect we would threat city, local, locality, and municipality as equivalent for collision-detection purposes. City/counties are slightly trickier, in that if one dataset calls SF a city and another calls it a county we'd be at risk for serving duplicate contests. On the bright side, we already have logic to treat aliased OCD IDs as belonging to every applicable district type, so if one of those datasets is BIP (which already uses OCD IDs) we would correctly detect the collision.

> Also, just to be clear, my original suggestion wasn't to invent a new identification scheme, but to introduce a parent object that would serve to group the districts within a numbering sequence. It is agnostic to the particular identification scheme: it doesn't require one but could still be used in conjunction with one (and in conjunction with 1622.2).

That either *is* an identifications scheme, with the slightly awkward extra restriction that different portions of the identifier go in different fields of different XML objects, or it doesn't solve any problems (because we can't tell whether two ElectoralDistricts in different feeds represent the same actual district or not).

Moreover, it is no more likely (read: 0% now, possible in the future) that any EMS will be able to supply this information usefully than that it will be able to supply OCD IDs for each electoral district, and the latter approach totally solves every problem that ElectoralDistrict.type is useful for anyway.

jdmgoogle:
I updated the enumerations so 'other' is always the last option (per the style guide).

If we come up with a concrete list of requests of suggestions or clarifications for 1622.2 DistrictType enumerations, I'll forward those along and then I'll feel comfortable merging this in. Suggestions so far seem to be:

1) Separate "Council" into "County Council" and "City Council" (are there "OtherCouncil" or is that not worth it/useful?); and
2) Please explain the difference between "local", "locality", "municipality", and "city".

Anything else?

jktomer:
LGTM given that we'll be passing the two questions above to 1622.

jdmgoogle:
Thanks. Merging now and forwarding comments to 1622.2 today or tomorrow.

Issue #116, opened 2015-04-21
Style: should type names end in "Type"?


cjerdonek:
I noticed that some type names end in "Type" (e.g. "ContestType" and "DetailAddressType"), but there doesn't seem to be a pattern as to which ones do.  For types like those, the suffix seems redundant to me.  Also, if such types end in "Type," it seems like they all should.

Personally, I think suffixes like "Type" and "Status" should be reserved for enumerations like "BallotMeasureType" and "CandidatePreElectionStatus," because they enumerate type _names_ (and not simply because they are the name of a type).  I think the style guide should clarify this and remove the ambiguity as to when the suffix should be present.

jdmgoogle:
I generally agree with this. The current naming also uses the FooType naming scheme when referring to abstract base classes which can never be instantiated. I'm okay with the FooType naming being used there, too.

cjerdonek:
If we go with a convention for abstract base classes, I'd prefer something distinct that we don't also use for some concrete classes.  In other words, if we use "FooType" for abstract base classes, I think Type should be disallowed for others (e.g. the BallotMeasureType enum of ballot-measure, initiative, referendum, etc).  Another option is to choose a different convention for abstract base classes like "FooBase".

jdmgoogle:
I'm okay with pure-abstract classes being renamed to "FooBase".

cjerdonek:
I created PR #121 for this.

cjerdonek:
Closed by PR #121.

PR #120, opened 2015-04-21
Venetian Blind conversion, first pass

pstenbjorn:
@jdmgoogle @jungshadow @cjerdonek @jktomer this is a first pass at converting the current version of the schema to Venetian Blind.  The only other changes I have made is to modify some of the underlying elements so they would validate.  This schema validates.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -15,7 +15,7 @@ reference elements for polling location
             <xs:all>
               <xs:element name="DateTime" type="xs:dateTime" />
               <xs:element name="Description" type="InternationalizedText" minOccurs="0" />
-              <xs:element name="FeedContactInformationId" type="ContactInformationType" minOccurs="0" />
+              <xs:element name="FeedContactInformation" type="ContactInformationType" minOccurs="0" />
               <xs:element name="Name" type="xs:string" />
               <xs:element name="OrganizationUrl" type="xs:anyURI" minOccurs="0" />
               <xs:element name="TermsOfUseUrl" type="xs:anyURI" minOccurs="0" />
@@ -76,7 +76,7 @@ reference elements for polling location
         <xs:element name="Locality" type="LocalityBase" />
         <xs:element name="PollingLocation" type="PollingLocationBase" />
         <xs:element name="Precinct" type="PrecinctBase" />
-        <xs:element name="Office" type="OfficeBase" />
+        <xs:element name="Office" type="Office" />
         <xs:element name="Party" type="PartyBase" />
         <xs:element name="PartyContest" type="PartyContestType" />
         <xs:element name="PartySelection" type="PartySelectionType" />
@@ -281,7 +281,7 @@ reference elements for polling location
     </xs:sequence>
   </xs:complexType>
 
-  <xs:complexType name="OfficeBase">
+  <xs:complexType name="Office">
     <xs:sequence>
       <xs:element name="ContactInformation" type="ContactInformationType" minOccurs="0" maxOccurs="unbounded" />
       <xs:element name="ElectoralDistrictId" type="xs:IDREF" />
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -5,621 +5,632 @@ added ocdid to district
 reference elements for polling location
 -->
 <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" version="5.0.1" xmlns:xs="http://www.w3.org/2001/XMLSchema">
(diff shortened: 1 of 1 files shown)


cjerdonek:
Thanks, @pstenbjorn.  Some comments on the PR:

1. It looks like you are being more aggressive with respect to type definitions than the style guide demands.  For example, the VipObject is getting a type definition even though it doesn't meet any of the three conditions.  Personally, I think we should only define a type if it meets one of the conditions, as those are the instances in which the type definitions are being "reused."

2. In issue #116, we are leaning towards not having type names end in "Type."  See the discussion there for reasons.

3. The style guide also says how the type definitions should be ordered (i.e. alphabetically by category of enums, simple types, then complex types).

4. The style guide has the root element going after the type definitions as opposed to before.

pstenbjorn:
@cjerdonek thank you for your quick attention to this, to address your questions.  

1.) This one style for Venetian Blind schemas - to have a names root element which has an associated abstract class for the schema for example this is the root of the 1622.2

```<xsd:element name="ElectionReport" type="ElectionReportType"/>```

But if we would like multiple root elements, I will be happy to expand this.

2.) I saw this discussion, I was unaware that #116 was now our official style but I will make sure we conform
3. and 4.) I can make modifications to they schema as agreed upon, sorry I missed correct ordering.  

Thanks

**Comment by cjerdonek** [comment]
Thanks, @pstenbjorn.  To your comments--

Regarding (1), in my mind I was still thinking one root element (which would be the VipObect).  It's just that it would be defined via an inlined / anonymous type as it is currently:

```xml
<xs:element name="VipObject">
  <xs:complexType>
    ...
```

as opposed to also having an associated "named" type definition as in your initial PR.  This would be because the VipObject type doesn't meet any of the criteria spelled out in the style guide for a type being "reused" (since the VipObject appears only once as an element and isn't referenced, etc).  I know this may not be considered a "pure" version of Venetian Blind, but the style is "rule based" and is in the spirit of Venetian Blind.  It just doesn't involve _every_ type having a named definition (only those being reused in the sense spelled out in the guide).

Regarding (2), #116 isn't our official style yet since it hasn't been added to the style guide.  But it appeared to me as if it was leaning that way.  So if you have an opinion on #116 (either yay or nay), feel free to weigh in.  I just wanted to point that out prior to committing name changes to every type.

jungshadow:
This looks like a good start, @pstenbjorn. One other administrative note: to the styleguide point, can you be sure to set your editor to 2 spaces rather than 4 for indents?

cjerdonek:
@pstenbjorn Before you start work on a second try at this, do you understand better which types are supposed to be global and which local?

Also, there's nothing wrong with spreading this PR over a number of commits as you work on it.  (Personally, I would prefer it in part because smaller commits are easier to review and comment on.)  Also, that way issues can be caught earlier and save you some time rather than propagating a change that may not be needed to the entire file.

pstenbjorn:
Just committed https://github.com/votinginfoproject/vip-specification/commit/c1639611ed449e5e42eea68631912b0529b822e8
This includes conversion to Venetian Blind style, and per @jdmgoogle alpha ordering elements and changing `ContactInformationId` references to inline `ContactInformationType` instances.

@cjerdonek also included all elements that aren't referenced in the base element for simplicity.

cjerdonek:
Hmm, I would have preferred this to be broken up into smaller changes that can be committed to the vip5 branch incrementally.  There is a lot to review here that seems to address multiple issues.  And it would have been easier to catch changes before they are propagated to the entire file (which is something I mentioned [here](https://github.com/votinginfoproject/vip-specification/pull/120#issuecomment-97470576).  For example, the "Base" suffix is getting applied to what looks like most types.

Also, are these commits made against the latest version of vip5?  I'm seeing old stuff in the diff.

pstenbjorn:
@cjerdonek I pulled fresh vip5 xsd file today to refactor.  I will be happy to break this into multiple PRs one that addresses just the Venetian Blind and then add PRs to address element ordering and `ContactInformation`.

cjerdonek:
Thanks, @pstenbjorn.  Yes, I think it would be easier to do only the Venetian blind portion (moving chunks from inlined types to named types as is), and then do things like reordering of elements, name changes, type changes, etc, as subsequent PR's.

While it will result in more PR's, I think it may actually have a good chance of being faster and with less chance of inadvertent errors, easier to review, etc.

cjerdonek:
Regarding pulling latest, the commit history shows this as starting from a version of vip5 from 10 days ago, which leaves out 20 or so commits.

pstenbjorn:
This has been replaced by #131

PR #121, opened 2015-04-21
Address issue #116: when type names should end in "Type"

cjerdonek:
Addresses issue #116.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -47,8 +47,17 @@ Markdown, etc).
   </xs:simpleType>
   ```
 
-* The names of abstract "base class" types should end in "Base" as a
-  reminder that the types should not be instantiated directly.
+* The names of abstract "base class" types (e.g. those with `abstract="true"`)
+  should end in "Base" as a reminder that such types should not be
+  instantiated directly.  For example--
+
+  ```xml
+  <xs:complexType name="ContestBase" abstract="true">
+    <xs:sequence>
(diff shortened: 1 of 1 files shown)


jdmgoogle:
Looks good to me. Thanks.

cjerdonek:
@pstenbjorn, @jungshadow approve?

pstenbjorn:
@cjerdonek a little more clarification about what you mean instantiated directly.  Can you provide examples about when you see the Base predicate used?  I thought I understood but in reading the .md file it isn't clear to me.

cjerdonek:
@pstenbjorn That was in response to a comment that @jdmgoogle made about abstract base classes.  If you look at 1622.2's spec, you can see some types marked `abstract="true"`: http://grouper.ieee.org/groups/1622/groups/2/WorkingDocuments/V42/1622_2-election_results_reportingV42.xsd

These are classes that aren't "instantiated directly," meaning that they are only used as base classes for other types that are instantiated.  In looking at VIP's spec, though, I see we don't carry over the `abstract="true"` notation, so perhaps VIP doesn't have abstract base classes after all.  Or maybe we do, but just don't notate them as such.

jdmgoogle:
@pstenbjorn Contest and BallotSelection should be abstract=true and renamed to ContestBase and BallotSelectionBase.

jungshadow:
This LGTM, but, to @pstenbjorn's point, I think maybe another example may help clarify things a bit more. That said, it doesn't hold things up for me.

PR #123, opened 2015-04-22
Update contributing documentation

jdmgoogle:
Update contributing documentation to include pre-PR validation steps and required approvals from VIP member organizations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ While the following information shouldn't be considered hard-and-fast rules, the
 3. Once it's done and tested, create a pull request to move it into the current working branch.
 4. At that point, some discussion might happen. In order to get approval for the pull request, you will need (at a minimum) approval from at least one representative from Pew and one representative from Google. However it is important to note that other members have substantial technical and election background as well, so please take all feedback to heart, regardless of the source.
     1. Google approvers: @jdmgoogle @jktomer
-    2. Pew approvers: @jungshadow @jen-tolentino
+    2. Pew approvers: @jungshadow @jen-tolentino (Pew employees) @pstenbjorn (official Pew proxy)
 5. When it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged into the current working branch by the developer who created the pull-request.
 6. Delete the feature/bug branch.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,8 @@ While the following information shouldn't be considered hard-and-fast rules, the
     2. If you have the `xmllint` tool on your system, please run `xmllint --postvalid --nonet --xinclude --noout --schema vip_spec.xsd sample_feed.xml`
 3. Once it's done and tested, create a pull request to move it into the current working branch.
 4. At that point, some discussion might happen. In order to get approval for the pull request, you will need (at a minimum) approval from at least one representative from Pew and one representative from Google. However it is important to note that other members have substantial technical and election background as well, so please take all feedback to heart, regardless of the source.
+    1. Google approvers: @jdmgoogle @jktomer
+    2. Pew approvers: @jungshadow @jen-tolentino
 5. When it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged into the current working branch by the developer who created the pull-request.
 6. Delete the feature/bug branch.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ While the following information shouldn't be considered hard-and-fast rules, the
     1. You can use http://www.utilities-online.info/xsdvalidation/ to do this validation online, or
     2. If you have the `xmllint` tool on your system, please run `xmllint --postvalid --nonet --xinclude --noout --schema vip_spec.xsd sample_feed.xml`
 3. Once it's done and tested, create a pull request to move it into the current working branch.
-4. At that point, some discussion might happen. In order to get approval for the pull request, you will need approval from two people, including one representative from Pew and one representative from Google (Pew and Google employees still need two approvers and cannot self-approve). However it is important to note that other members have substantial technical and election background as well, so please take all feedback to heart, regardless of the source.
+4. At that point, some discussion might happen. In order to get approval for the pull request, you will need approval from two people, including one representative from Pew and one representative from Google (Pew and Google employees still need two approvers and cannot self-approve, but it is not required that the second approver be from the organization of the PR author). However it is important to note that other members have substantial technical and election background as well, so please take all feedback to heart, regardless of the source.
     1. Google approvers: @jdmgoogle @jktomer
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,8 +15,12 @@ While the following information shouldn't be considered hard-and-fast rules, the
 
 1. Create a branch to work on a fix/feature (a fix/feature should have a companion bug/enhancement issue). Start the branch with either "feature/..." or "bug/...".
     1. If you're not a member of the VIP Spec team, fork the repository and follow the same process.
-2. Once it's done and tested, create a pull request to move it into the current working branch.
-3. At that point, some discussion might happen and, when it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged into the current working branch by the developer who created the pull-request.
-4. Delete the feature/bug branch.
+2. Before sending out a pull request, please make sure the resulting XSD and sample feed XML still validate.
+    1. You can use http://www.utilities-online.info/xsdvalidation/ to do this validation online, or
+    2. If you have the `xmllint` tool on your system, please run `xmllint --postvalid --nonet --xinclude --noout --schema vip_spec.xsd sample_feed.xml`
```
cjerdonek: For convenience, might be nice to have a simple shell script in the repo containing this command.

CONTRIBUTING.md
```diff
@@ -15,8 +15,12 @@ While the following information shouldn't be considered hard-and-fast rules, the
...
+5. When it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged into the current working branch by the developer who created the pull-request.
+6. Delete the feature/bug branch.
 
-At any one point in time ("feature/" and "bug/" temporary branches aside), there should only be a dev branch (called 'vip5' in the vip-specification's case, but it may change to simply 'dev' in the future), a stable branch (called 'master' in most cases), and, if necessary, a documentation branch (called 'gh-pages' if hosted on GitHub).
\ No newline at end of file
+At any one point in time ("feature/" and "bug/" temporary branches aside), there should only be a dev branch (called 'vip5' in the vip-specification's case, but it may change to simply 'dev' in the future), a stable branch (called 'master' in most cases), and, if necessary, a documentation branch (called 'gh-pages' if hosted on GitHub).
```
cjerdonek: Out of curiosity, what changed between these two lines?  (I don't see the difference.)  Might be worth wrapping to reasonable line length in this document.

CONTRIBUTING.md
jdmgoogle: I'd thought about that, but I'm not sure how many people are using *nix and how many are using Windows (or any other platform which might not support bash).

CONTRIBUTING.md
jdmgoogle: Apparently some end-of-line stuff. This might need to be cleaned up in general.

CONTRIBUTING.md
cjerdonek: I think it's worth adding a *nix only script.  (FWIW, I seem to have xmllint on my Mac OS X system.)  I also think it would be fine to add a similar Windows-specific script if there is one (I've seen projects with parallel `*.sh` and `*.bat` files to address this issue, provided the scripts are simple (e.g. wrappers).

cjerdonek:
I meant to add as an overall comment: for transparency and convenience, I think it's important that the documentation list the usernames of the people from each required group (e.g. like I had asked in the issue thread).  This way, for example, people know who they have to bug, when they already have approval, etc.

CONTRIBUTING.md
jungshadow: I know we have a few people developing on Windows (@pstenbjorn for one). Rather than try to have a single-platform script (or try to develop a solution for all), I think the command line example is sufficient.

jungshadow:
@cjerdonek Sounds good. We (Pew) will have an answer about the "who" by the end of today.

jungshadow:
Beyond that, LGTM

jdmgoogle:
@cjerdonek I'll update the documentation once I get the list from Pew.

CONTRIBUTING.md
jktomer: Different editors have different opinions about whether the last line in a file should have a trailing newline or not. People generally ignore diffs in that particular aspect, because no actual human ever cares.

jdmgoogle:
Documentation has been updated to reflect that @pstenbjorn is an official Pew proxy.

jungshadow:
My "LGTM" still stands :)

jdmgoogle:
Once I get a second LGTM I'll merge.

jungshadow:
@jdmgoogle Uh…you can't self-approve with the new guidelines? ;)

jdmgoogle:
I should probably update the documentation to say that each PR still needs two non-author approvers, even if that person is from Pew or Google. :) In which case I can't self-approve.

cjerdonek:
That occurred to me, too.  To clarify: in the case of a Google employee, does the second approver need to be another Google employee, or can it be anyone?

jdmgoogle:
Darn edge cases. :)

There are two ways to look at it, and this seems like a "for the community" question:

1) The "Google approval" is implicit in them putting it forward, and it can come from anyone; or
2) When putting forth a PR, affiliation is ignored and the author still has to get approval from one person from each org.

I'm waffling on this and I'm interested in everyone's thoughts.

cjerdonek:
Here's one possible way of thinking about it. The original proposal served two purposes: (1) ensures buy-in from both groups (Google and Pew), and (2) ensures that at least two other pairs of eyes looked at it (which is an orthogonal, good practice in general).  So I would lean towards the first option.

cjerdonek:
Incidentally, the second option would also have practical issues, because Google only has two employees.  A Google employee would be out of luck if the other person on vacation.  Another way of looking at this option: a Google employee would ironically have a "harder" time getting something approved because they would only have one Google employee to choose from whereas others have two. That doesn't seem right. :grinning:

jdmgoogle:
Pinging @jktomer @jungshadow @pstenbjorn @nomadaisy @pkoms @demcg @MariaDemocracyWorks etc for feedback.

jktomer:
Traditionally there's a distinction between "minimum number of reviewers" and "minimal set of reviewer-or-author qualifications." So we could say:

1. Every pull request needs at least one LGTM from (one or more) reviewers
2. At least one of the author or reviewer must be from Google
3. At least one of the author or reviewer must be from Pew

Or we could change rule 1 to say you need *two* reviewer LGTMs, in which case a Google- or Pew-proposed pull request would need one LGTM from a person from the other organization, and one additional LGTM from anyone.

Personally, I think we've gotten close enough to perfect on this score that the cost of getting better exceeds its value, and any answer to these questions is fine.

jungshadow:
I like @jktomer's take on this (i.e. the 3-steps) and I agree that trying to hit all of the edge cases is more work than it's worth.

jdmgoogle:
Okay, feedback has been incorporated and the text matches the discussion. Can I get the right set of LGTMs to start this off right? :)

jktomer:
LGTM

jungshadow:
:+1:

PR #124, opened 2015-04-22
Sync spec to latest revision of the 1622.2 standard

jdmgoogle:
Release candidate of the 1622.2 standard can be found at:
http://grouper.ieee.org/groups/1622/groups/2/WorkingDocuments/V42/1622_2-election_results_reportingV42.xsd

Changes include:
- Renaming Contact to ContactInformation;
- Creating a TimeWithZone pattern to reuse in the Hours object; and
- Renaming a few attributes, enums, and elements to be consistent with 1622.2.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -18,7 +18,7 @@ reference elements for polling location
               <xs:element name="DateTime" type="xs:dateTime" />
               <xs:element name="Description" type="InternationalizedText" minOccurs="0" />
               <xs:element name="OrganizationUrl" type="xs:anyURI" minOccurs="0" />
-              <xs:element name="FeedContactId" type="xs:IDREF" minOccurs="0" />
+              <xs:element name="FeedContactInformationId" type="xs:IDREF" minOccurs="0" />
               <xs:element name="TermsOfUseUrl" type="xs:anyURI" minOccurs="0" />
             </xs:all>
             <xs:attribute name="id" type="xs:ID" use="required" />
@@ -124,7 +124,7 @@ reference elements for polling location
               <xs:element name="Department" maxOccurs="unbounded">
                 <xs:complexType>
                   <xs:all>
-                    <xs:element name="ContactId" type="xs:IDREF" minOccurs="0" />
+                    <xs:element name="ContactInformationId" type="xs:IDREF" minOccurs="0" />
                     <xs:element name="ElectionOfficialPersonId" type="xs:IDREF" minOccurs="0" />
                     <xs:element name="VoterService" minOccurs="0" maxOccurs="unbounded">
                       <xs:complexType>
@@ -136,10 +136,10 @@ reference elements for polling location
                             The contact information below can be used to override or
                             add specific fields to the base Departmental contact information if
                             the service has different information.  For example, if the voter
-                            service has its own phone number, the Contact object below can be
-                            an object containing only a Phone element.
+                            service has its own phone number, the ContactInformation object
+                            belowe can be an object containing only a Phone element.
                           -->
-                          <xs:element name="ContactId" type="xs:IDREF" minOccurs="0" />
+                          <xs:element name="ContactInformationId" type="xs:IDREF" minOccurs="0" />
                           <!--
                             This is for use if a certain person handles the particular service,
                             for example a contact person for overseas voting.
@@ -260,7 +260,7 @@ reference elements for polling location
           </xs:complexType>
         </xs:element>
 
-        <xs:element name="Contact">
+        <xs:element name="ContactInformation">
           <xs:complexType>
             <xs:sequence>
               <xs:element name="Uri" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
@@ -289,7 +289,7 @@ reference elements for polling location
               <xs:element name="IsPartisan" type="xs:boolean" minOccurs="0" />
               <xs:element name="FilingDeadline" type="xs:dateTime" minOccurs="0" />
               <xs:element name="Term" type="Term" minOccurs="0" />
-              <xs:element name="ContactId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
+              <xs:element name="ContactInformationId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
             </xs:sequence>
             <xs:attribute name="id" type="xs:ID" use="required" />
           </xs:complexType>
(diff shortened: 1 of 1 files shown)


jdmgoogle:
There are several SURVEY comments currently embedded in this PR, so please look them over and provide feedback.

Cc @jktomer @jungshadow @pstenbjorn @cjerdonek @nomadaisy @pkoms @demcg

vip_spec.xsd
```diff
@@ -136,10 +136,10 @@ reference elements for polling location
                             The contact information below can be used to override or
                             add specific fields to the base Departmental contact information if
                             the service has different information.  For example, if the voter
-                            service has its own phone number, the Contact object below can be
-                            an object containing only a Phone element.
+                            service has its own phone number, the ContactInformation object
+                            belowe can be an object containing only a Phone element.
```
cjerdonek: typo

vip_spec.xsd
```diff
@@ -337,8 +340,11 @@ reference elements for polling location
         <xs:element name="Person">
           <xs:complexType>
             <xs:sequence>
+              <!-- SURVEY: Types xs:string and InternationalizedText seem
+                   to be used inconsistently here. Suggestions?
```
cjerdonek: I think I have an answer, but just wanted you to clarify first: which inconsistency are you referring to?

vip_spec.xsd
```diff
@@ -511,13 +507,13 @@ reference elements for polling location
 
     could be:
 
-      <Office textIdentifier="office_mayor">
+      <Office identifier="office_mayor">
         <text language="en">Mayor</text>
         <text language="es">Alcalde</text>
         <text language="zh">市長</text>
```
cjerdonek: I know this isn't part of the PR, but could you fix the casing of the `<Text>` tags here?

vip_spec.xsd
jdmgoogle: I'm just wondering why profession and title are internationalized but names, nicknames, etc, aren't.

pstenbjorn:
@jdmgoogle this looks good and is similar to the the changes made in #120.

vip_spec.xsd
cjerdonek: My take on this is that if it's info that someone submitted to file (e.g. their personal name), then that should be a string simply providing what they filled out as is.  In contrast, the "BallotName" for the candidate is the name as it should appear on the ballot, and so can have translations.  Profession is a bit of an exception in that it can appear translated on the ballot (see picture below).  When I first saw that, I was thinking there should also be a "BallotProfession" as distinct from the profession listed in the Person object.

sample

vip_spec.xsd
```diff
@@ -298,6 +298,9 @@ reference elements for polling location
         <xs:element name="Party">
           <xs:complexType>
             <xs:sequence>
+              <!-- SURVEY: 1622.2 has abbreviation as a string, not InternationalizedText.
+                   Which makes more sense?
```
cjerdonek: In my mind, it all depends on whether non-English ballots display abbreviations in languages other than English.  I don't have access to current sample ballots, though, so I can't provide any data points.

vip_spec.xsd
cjerdonek: Just to make my suggestion concrete, I think all fields in the Person object should be xs:string, but then "BallotProfession" should be added as InternationalizedText to the Candidate object.  I'm making this suggestion under the assumption that "BallotName" in the Candidate object means the name of the candidate as it should appear on the ballot.

vip_spec.xsd
jdmgoogle: Fixed.

vip_spec.xsd
jdmgoogle: Fixed.

vip_spec.xsd
jdmgoogle: Currently VIP And 1622.2 do this differently, so absent any compelling use case I'm inclined to just go the 1622.2 compatibility route just to make it easy on election officials.

vip_spec.xsd
jdmgoogle: Hrm, interesting. In that case I actually motion that was keep it as it is (i.e., the same as 1622.2) since the information will still be there and in the same format as 1622.2. We can add it to the list of suggestions for 1622.2 if there's a brief public comment period. Otherwise I can live with it as-is.

cjerdonek: I thought about raising that with 1622.2, but as a latecomer, and in getting the impression that they are really trying to wrap up the process and only want serious issues, I didn't saying anything.

cjerdonek:
LOTM

jdmgoogle:
I've updated a few things based on the conversations here. There's still one last SURVEY question left, probably best answered by one of @pstenbjorn, @jungshadow, or @jktomer.

Please take another look and approve/provide feedback before I merge. Thank you.

pstenbjorn:
@jdmgoogle in regard to the use of the `InternationalizedText` element, I would favor having a rule that defines elements that are of `xs:string` where there are known instances of official translation text required under state or federal law to be cast as `InternationalizedText` otherwise use `xs:string`.  `Abbreviation` IMO should be a `xs:string`.

In regard to `VotesWithCertification` and '=`CertificationEnum'=` should be removable.

vip_spec.xsd
```diff
@@ -416,13 +404,14 @@ reference elements for polling location
...
-      <xs:element name="Name" type="InternationalizedText" />
-      <xs:element name="Abbreviation" type="InternationalizedText" minOccurs="0" />
+      <xs:element name="Name" type="xs:string" />
+      <xs:element name="Abbreviation" type="xs:string" minOccurs="0" />
       <xs:element name="BallotTitle" type="InternationalizedText" minOccurs="0" />
       <xs:element name="BallotSubTitle" type="InternationalizedText" minOccurs="0" />
       <xs:element name="HasRotation" type="xs:boolean" minOccurs="0" />
       <xs:element name="VoteVariationType" type="VoteVariation" minOccurs="0" />
       <xs:element name="OtherVoteVariationType" type="xs:string" minOccurs="0" />
+      <!-- SURVEY: What is the use case for this field? It's empty in the sample feed. -->
```
jktomer: From the equivalent field in the 3.0 docs: "The electorate_specifications field specifies any changes to the eligible electorate for this contest past the usual, "all registered voters" electorate. This subtag will most often be used for primaries and local elections. In primaries, voters may have to be registered as a specific party to vote, or there may be special rules for which ballot a voter can pull. In some local elections, non-citizens can vote."

e.g. "registered members of Party X only" for primaries.

jktomer:
LGTM

jdmgoogle:
@pstenbjorn Thanks. Abbreviation is now an xs:string. We should update the documentation to reflect this policy.

Merging now.

PR #125, opened 2015-04-24
Incorporate small changes to CSV language in documentation

nomadaisy:
Changes requested from @jen-tolentino to language around CSV documentation

### DIFF
--- a/index.html
+++ b/index.html
@@ -61,9 +61,9 @@
           <h1 class="title"><abbr title="Voting Information Project">VIP</abbr> Format Specification <span class="version">Version 3.0</span></h1>
 
           <h2>General Introduction</h2>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -74,8 +74,8 @@ <h1 class="title"><abbr title="Voting Information Project">VIP</abbr> Format Spe
...
-          <p>If your office is producing flat files instead of an XML file, the files must be comma-delimited .txt files. While the format of the files differs slightly from the XML, the attributes included must conform to the field specifications below.<p/><p/> Examples of the flat files are below each XML example, along with the files included to link those elements when the files are compiled into an XML feed. To read more about the flat files, see complete file header templates and look at example files, you can do so in this <a href="https://github.com/votinginfoproject/csv-templates">GitHub repository.</a> </p>
+          <h3>Comma-delimited flat files (CSV)</h3>
+          <p>If your office is producing flat files, they must be comma-delimited .txt files. The CSV column names differ slightly from XML tag names.<p/><p/> Examples of the flat files are below each XML example and include the necessary files to link those elements together as needed. To read more about the flat files, view complete file header templates, and review example files, you can do so in this <a href="https://github.com/votinginfoproject/csv-templates">GitHub repository.</a> </p>
```
jungshadow: @nomadaisy Random aside (and this might actually be a question for @tie-rack), but does the filename have to end in .txt? Could it also be .csv? This is more about our intake process and how hardened it is. Since there really is no CSV standard, I can see arguing either way (i.e. either that we accept both or mandate only one).

jungshadow:
@nomadaisy One quick note is just around the process of submitting PRs. If you can create a 'feature/' branch and do a PR against master rather than committing directly to master, that would make me sleep easier :) We can always rebase master on gh-pages, since no one looks to gh-pages as the canonical source.

Other than the process note and the line question, this looks good. Thanks!

jungshadow:
@nomadaisy *bump*

nomadaisy:
Great, thanks. Process changes noted, I was able to read up on the updates from last week. I'll go ahead and merge this with gh-pages and use the new process going forward.

index.html
nomadaisy: Currently, they must be .txt files. This could change with the new Data QA Engine, and ideally, it would accept both.

Issue #94, opened 2015-04-09
Linking Precinct object to BallotStyle


Josh-LACRRCC:
Since the restructuring of the schema to better follow 1622.2, I no longer see a method of connecting a Precinct to a BallotStyle.

The Precinct's "BallotStyleImageUrl" function appears be taken care of by the BallotStyle's "InfoUrl".
The "BallotStyleImageUrl" element could then be replaced by

```
<xs:element name="BallotStyleID" type="xs:IDREF" minOccurs="1" maxOccurs="1" />
```

or be added following if the image url needs to remain in that location.

Other ideas?

jungshadow:
@Josh-LACRRCC Did this question get answered yesterday at the spec clean-up session? I want to make sure this didn't fall through the cracks. /cc @pstenbjorn @jdmgoogle

Josh-LACRRCC:
@jungshadow Yes. Revisit after further 1622.2 integration.

jdmgoogle:
I sat down and discussed this with @jktomer, and here's the situation: right now we (Google) figure out the ballot contests by (a) getting the list of electoral districts associated with a precinct from the Precinct object, (b) getting the list of contests associated with an electoral district from the Contest object, and (c) doing a join on those two lists. These linkages are all still present in the most recent draft of VIP 5. However there are also BallotStyle objects in the VIP 5 spec, but currently they are not referenced by anything. There are four options we considered:

1. Leaving this as-is;
2. Removing the BallotStyle (and OrderedContest) objects, and continuing to get the ballot contents by joining on ElectoralDistrict (the current method);
3. Keep ElectoralDistrict in Precinct, and add a reference to a BallotStyleId to Precinct as well; or
4. Remove ElectoralDistrict from Precinct, and add BallotStyleId to Precinct.

We shot down (1) as being pointless since BallotStyle and OrderedContest would be of no use and just take up space in the feed.

We would prefer to not remove ElectoralDistrict from Precincts, since this information is useful in many ways and improves the search experience (this removes option (4)).

We suspect that removing BallotStyle and OrderedContest from the schema would be less-than-desirable since it's a superior and more direct way to indicate ballot layout than our join-through-ElectoralDistrict heuristic (this removes option (2)).

Now we're left with option 3) Keep ElectoralDistrict in Precinct and add a BallotStyleId field. If such a field is present then we'd use the referenced BallotStyle, but if it isn't then we can fall back on our joining heuristic.

Comments?

[...]

PR #127, opened 2015-04-27
Replace BallotStyleImageUri with a BallotStyleId

jdmgoogle:
The BallotStyle object can hold both an image URI as well as a set of contests (i.e., it's a superset of the URI, without loss of flexibility).

This should address issue #94.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -87,7 +87,7 @@ reference elements for polling location
               <xs:element name="Ward" type="xs:string" minOccurs="0" maxOccurs="1" />
               <xs:element name="PollingLocationId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
               <xs:element name="IsMailOnly" type="xs:boolean" minOccurs="0" maxOccurs="1" />
-              <xs:element name="BallotStyleImageUrl" type="xs:anyURI" minOccurs="0" />
+              <xs:element name="BallotStyleId" type="xs:IDREF" minOccurs="1" maxOccurs="1" />
               <xs:element name="ExternalIdentifiers" type="ExternalIdentifiers" minOccurs="0" />
             </xs:sequence>
             <xs:attribute name="id" type="xs:ID" use="required" />


jktomer:
LGTM

Should we have a todo to update the documentation with a note about a precinct's ballot style being required to cover the exact set of contests inferrable from its electoral districts?

jdmgoogle:
@jktomer Possibly. Is that a Google-specific implementation detail, though, or something we (VIP) want feed creators to keep in mind?

jktomer:
The latter, in my opinion (there's semantic ambiguity otherwise, which we've done our very best to avoid).

jungshadow:
LGTM

cjerdonek:
It doesn't seem to me like the ballot style object should be _required_ to cover every contest.  Semantically, I think the ballot style object should only be used for ordering contests, and that the actual contests would be obtained the usual way.  Also, there are cases where the ballot style object might only have partial information (e.g. only an URL, or only covering some contests, e.g. federal and state but not local), so I don't think the ballot style object should be disallowed in such cases.

jdmgoogle:
@cjerdonek I don't think there's anything in the spec which *requires* feed creators to cover every contest. Obviously it's preferable if they do, but right now the `BallotStyle` object just provides a way to order contests.

cjerdonek:
@jdmgoogle I was responding to @jktomer's question above as to whether it should be made required:

> Should we have a todo to update the documentation with a note about a precinct's ballot style being required to cover the exact set of contests inferrable from its electoral districts?

jdmgoogle:
@cjerdonek Ah, okay. Sorry for the confusion.

And what @jktomer was saying is that if a Contest **C** is associated with ElectoralDistrict **A** and a precinct is associated with ElectoralDistrict **A** and BallotStyle **S** then BallotStyle **S** should be required to reference Contest **C**.

cjerdonek:
Yes, I was saying I don't think we should require that.

jdmgoogle:
I don't know if there's any way we (VIP) can enforce that. But we (Google) would view a feed that listed one set of `Contests` in the `BallotStyle` object (and referenced that `BallotStyle` object from the `Precinct` object) but had a different set of `Contests` tied through `ElectoralDistrict` objects to a `Precinct` as being an incomplete or inconsistent feed, and would provide no guarantees as to what actually gets presented to the end user.

cjerdonek:
One suggestion for you / Google is to consider guaranteeing that the set of contests is always obtained through the electoral districts.  Otherwise, it can penalize / disincentivize someone for trying to provide ballot styles but not doing it quite right (e.g. if providing an incorrect ballot style can potentially invalidate the contests displayed to the user).  In other words, I'm suggesting giving priority to the data via electoral districts over the ballot style (since it is just ordering information which is a nice-to-have rather than a requirement).

jdmgoogle:
The problem is that if `BallotStyles`  ~> `Contest` and the `Precinct` ~> `ElectoralDistrict` ~> `Contest` links are not equivalent then we are getting conflicting data from the provider, in which case we can't provide any guarantees about how it will be processed. If a provider feels they may not be able to populate the `BallotStyle` linkages correctly, the answer is to provide *only* linkages through the `ElectoralDistrict` process as opposed to including contradictory data in the feed.

cjerdonek:
It was just a suggestion.  What you guarantee is up to you (i.e. you can provide a guarantee in this situation if you want to).  One of the examples I gave in my previous comment is if ordering information is provided for a subset of contests:

> Also, there are cases where the ballot style object might only have partial information (e.g. only an URL, or only covering some contests, e.g. federal and state but not local),

In any case, I was just trying to help and don't have a stake in which way you go.

kennethmbennett:
??Hi all,


I know I don't chime in much (I do follow the posts), but wanted to weigh in on this one.  Any solution that does not support the display of candidates in the appropriate rotated order (per California Elections Code) will constrain the usefulness of the feed for California election administrators, since "official" data coming from election administrators arguably should comply with California law.  I believe there are other states with candidate rotations requirements as well (.e.g. Arizona).


My understanding is that candidate rotation is supported in the schema through the BallotStyle object.  If that is true, then I would want some assurance that candidate information could be provided in rotated order through that path.


If am misunderstood, please ignore this post.  But my general comment remains, please do whatever it takes to ensure that the schema supports candidate rotation by ballot style.


Thanks,

Kenneth

________________________________

jdmgoogle:
Kenneth,

You are correct in that the feed spec supports ballot rotation, both through the `BallotStyle` element and the `Contest` element.

What does California law say about rotation on (a) sample ballots mailed to voters, and (b) any online sample ballot lookup tools?

Ccing @jktomer as a heads-up.

johnpwack:
I know I don't chime in much either but also wanted to weigh in. I can't
speak for how candidate rotation is handled in VIP, as I haven't been
following this list closely enough. The discussions in the 1622.2 working
group on this issue have been kind of interesting. So, in 1622.2, Contest
and Candidate both have a sequence order associated with them. I've had
pushback from various people about these attributes, as the EMS vendors are
saying that they don't ever export any sequence numbers.  They are saying
that the only way to preserve rotation is to keep a record of the ballot
style per precinct/split, so that is one of the reasons why 1622.2 has the
BallotStyle element. The expectation, then, is that the contests and
candidates in the BallotStyle element will be written out in the correct
rotation.

This seems to make sense to me – I can't think of any other way to do it
unless one wants to calculate from the vote variation what the order is per
precinct/split, and this would probably not be a good thing to do.  At the
same time, I'm not a fan of deriving the ballot ordering necessarily – in
other words, the order of the ballot has to be derived from the order of
the XML elements.  I have pressed for sequence numbers to be added to the
BallotStyle element and I think this would be a better thing overall,
because I view the 1622.2 export as a simple base export format that other
applications can use, and what if they don't preserve the same ordering? I
guess there are pros and cons but I think providing a sequence number in
the BallotStyle is a safer thing to do.

As regards the existing sequence numbers in 1622.2 that are associated with
Candidate and Contest, I think they will be used more for the purpose of
displaying election results – for example, on a ticket, ordering the
candidates according to how they should be displayed.

My $.02, John

*---*

*John P. Wack*

johnpwack@gmail.com

301.640.6626 - cell

301.975.3411 - office

cjerdonek:
> What does California law say about rotation on (a) sample ballots mailed to voters, and (b) any online sample ballot lookup tools?

@jdmgoogle, CA Elections Code [Sec. 13300](http://www.leginfo.ca.gov/cgi-bin/displaycode?section=elec&group=13001-14000&file=13300-13317) says this--

    13300.  (a) ...
       (b) The sample ballots shall be identical to the official ballots,
    except as otherwise provided by law. The sample ballots shall be
    printed on paper of a different texture from the paper to be used for
    the official ballot.
    ...
    13300.7.  Notwithstanding any other provision of law, county and
    city elections officials may establish procedures designed to permit
    a voter to opt out of receiving his or her sample ballot, voter
    pamphlet, notice of polling place, and associated materials by mail,
    and instead obtain them electronically via e-mail or by accessing
    them on the county's or city's Internet Web site, provided that all
    of the following conditions are met:

As a side note, [Sec. 13109](http://www.leginfo.ca.gov/cgi-bin/displaycode?section=elec&group=13001-14000&file=13100-13121) speaks to the order of appearance of offices (as well as candidates in later sections)--

    13109.  The order of precedence of offices on the ballot shall be as
    listed below for those offices and measures that apply to the
    election for which this ballot is provided. Beginning in the column
    to the left:
       (a) Under the heading, PRESIDENT AND VICE PRESIDENT:
       Nominees of the qualified political parties and independent
    nominees for President and Vice President.
       (b) Under the heading, PRESIDENT OF THE UNITED STATES:
       (1) Names of the presidential candidates to whom the delegates are
    pledged.
       (2) Names of the chairpersons of unpledged delegations.
       (c) Under the heading, STATE:
       (1) Governor.
       ...

jdmgoogle:
@cjerdonek @kennethmbennett Having read over the section of CA law, here are my impressions:

1) The law discusses the sample ballot and how it _must_ be presented to the voter. Clearly this is a concern for both the state and the counties in how whatever lookup tool they use _presents_ the data to the voter.

2) _If_ this covers sample ballot lookup tools, _then_ CA, Pew, and Google will need to discuss this. I'm skeptical that any solution which is NOT an explicit `BallotStyle` based linkage will be able to cover the full ordering as outlined in Sec 13109. There's simply no way that Google can write code to _heuristically_ determine ballot ordering in a way that complies with California law _and_ any other state law which specifies detailed ballot ordering.

3) Any non-official lookup tool -- i.e., one not being embedded on an official site or operating as a representative of a state government agency -- is outside the scope of this discussion.

jktomer:
> They are saying that the only way to preserve rotation is to keep a record of the ballot style per precinct/split, so that is one of the reasons why 1622.2 has the BallotStyle element. The expectation, then, is that the contests and candidates in the BallotStyle element will be written out in the correct rotation. This seems to make sense to me – I can't think of any other way to do it
unless one wants to calculate from the vote variation what the order is per
precinct/split, and this would probably not be a good thing to do.

I think "probably" is a bit of an understatement :) The degree of variation in rules about ballot ordering is _extreme_ and some of them are very complicated, so I wouldn't trust anyone other than election officials (not even Google! :)) to calculate it from scratch.

>  At the same time, I'm not a fan of deriving the ballot ordering necessarily – in other words, the order of the ballot has to be derived from the order of the XML elements.  I have pressed for sequence numbers to be added to the BallotStyle element and I think this would be a better thing overall, because I view the 1622.2 export as a simple base export format that other applications can use, and what if they don't preserve the same ordering? I guess there are pros and cons but I think providing a sequence number in the BallotStyle is a safer thing to do.

One thing I dislike about attaching sequence numbers to the candidate and contest objects themselves is that it sets up a confusing situation where BallotStyles have an order that disagrees with the order implied by the sequence numbers on the referenced objects. In light of your explanation below, that now makes sense to me, but I can see it causing confusion for data suppliers and consumers without careful explanatory documentation.

As for using inferred sequence order vs. creating a new type that's effectively pair<integer, reference-to-object> for explicitly ordering things in sequences, I wouldn't necessarily oppose that but it seems more verbose than necessary, since I think every proposed underlying format for 1622 data supports ordered sequences natively.

> As regards the existing sequence numbers in 1622.2 that are associated with Candidate and Contest, I think they will be used more for the purpose of displaying election results – for example, on a ticket, ordering the candidates according to how they should be displayed.

That would make sense in that there can be a defined order in which candidates will be presented on a _results_ list even if they are presented in different orders on the ballots of different precincts. In that case I would propose dropping the sequence order elements from VIP's versions of these objects (since VIP doesn't cover results, these would only add confusion).

kennethmbennett:
Thanks for the CEC extract. Chris.  That’s helpful.  It’s the “sample ballots shall be identical to the official ballots” language that is most critical for election administrators, from the standpoint of being the provider of official ballot information.  If the official ballot has contests that are subject to candidate rotation rules, then the official sample ballot provided by the election jurisdiction, whether paper or electronic, needs to reflect that same ordering.

jktomer:
> One suggestion for you / Google is to consider guaranteeing that the set of contests is always obtained through the electoral districts. Otherwise, it can penalize / disincentivize someone for trying to provide ballot styles but not quite doing it right (e.g. if providing an incorrect ballot style can potentially invalidate the contests displayed to the user). In other words, I'm suggesting giving priority to the data via electoral districts over the ballot style (since the latter is just ordering information which is a nice-to-have rather than a requirement).

I'm not sure I agree: the primary purpose of supporting per-precinct ordering at all is so that people creating *official* tools can guarantee that the ballots are presented *exactly* as they need (as in the California example being discussed in this thread), it's important that we not create accidental violations of these expectations when there is questionable input data. Instead our focus is on providing clear, easy-to-understand reports of invalid data to election officials so they can fix their feeds.

> Also, there are cases where the ballot style object might only have partial information (e.g. only an URL, or only covering some contests, e.g. federal and state but not local), so I don't think the ballot style object should be disallowed in such cases.

Reconciling the partial order implied by one dataset (or portion of a dataset) with any other ordering and getting always-reasonable results is an intractable problem; I tried and failed to do this last year.

cjerdonek:
> The expectation, then, is that the contests and candidates in the BallotStyle element will be written out in the correct rotation.  This seems to make sense to me – I can't think of any other way to do it unless one wants to calculate from the vote variation what the order is per precinct/split, and this would probably not be a good thing to do.

@johnpwack, In the case of simple rotation, another option (which I'm not necessarily advocating) would be to provide only the first candidate in the rotation (e.g. an optional `RotationStartCandidateId` element).  I wouldn't be surprised if that alone would handle a great majority of cases.  Are there any jurisdictions that do permutations other than rotations across ballot styles?

jktomer:
>  In the case of simple rotation, another option (which I'm not necessarily advocating) would be to provide only the first candidate in the rotation (e.g. an optional RotationStartCandidateId element). I wouldn't be surprised if that alone would handle a great majority of cases. Are there any jurisdictions that do permutations other than rotations across ballot styles?

I strongly recommend against getting into the business of encoding state and local laws in the schema. Specifying a ballot order per precinct already makes a lot of extra work for consumers of this data. Providing an alternate mechanism to specify the same information then requires both producers and consumers to write more code (with more bugs!) for zero benefit (except, perhaps, marginally smaller file sizes before compression).

jdmgoogle:
I agree with @jktomer; there should be one way to specify this data. The only reason we also have the `ElectoralDistrict` path is because it's an artifact of the need to provide complete information about a contest. The current schema allows feed creators to specify ballot styles as a sub-precinct level (i.e., precinct splits). Additional optimizations which are tailored to a single jurisdiction -- even one as large as California -- but are already possible to represent with existing schema elements is just going to lead to unnecessary complexity, more bugs, and unpredictable user-facing behavior.

cjerdonek:
> Providing an alternate mechanism to specify the same information then requires both producers and consumers to write more code (with more bugs!) for zero benefit (except, perhaps, marginally smaller file sizes before compression).

I wasn't advocating it but merely pointing it out.  You said there was zero benefit, but one benefit of a more DRY approach is that it eliminates more possibilities of having an internally inconsistent feed (since otherwise candidate lists are being repeated, etc).  Out of curiosity, do we actually know of any jurisdiction that does something different from rotating a master list (i.e. are arbitrary permutations actually needed)?

jdmgoogle:
@cjerdonek Arizona randomly rotates the candidate list for each ballot.

cjerdonek:
@jdmgoogle Anything _other_ than rotating the candidate list.

jdmgoogle:
Sorry. To clarify, in Arizona one person may get ballot *A-B-C-D*, the next may get *B-A-C-D*, the next may get *D-A-C-B*, etc, etc; i.e., "pick candidate names out of a hat until you've got the full slate."

pstenbjorn:
@jdmgoogle @kennethmbennett @cjerdonek I just wanted to chime in and state the following that might be self evident:

True ballot rotations (e.g. Arizona) cannot be represented in sample ballots since there is an element of randomness to the rotation scheme (there is not one ballot style per precinct per election event).  In Los Angeles (and in Ohio, Indiana and elsewhere) the definition of "rotation" represents different ballot styles by precinct where the candidate sequence differs predictably from precinct to precinct.  

Given that the schema is designed to support ballot styles produced by local election officials it seems like a worthwhile pursuit to represent this sequencing in the schema.  In previous attempts at the 5.0 schema, we were able, albeit inelegantly, to reproduce this ballot rotation through instances of a ballot style element associated with precincts.

jungshadow:
@kennethmbennett @johnpwack Thanks for both watching and chiming in here. Always appreciate your input. 

@pstenbjorn Just to clarify, the `BallotStyle` element is linked to a `Precinct` element via `BallotStyleId` and this can represent the candidate and contest ordering in places like OH, IN, and CA. Correct? If that's the case and we can serve the immediate requirements of most states, than this gets another "LGTM" from me (assuming no other issues).

Agree with @jktomer and @jdmgoogle that we can't encode individual state laws in the schema. I can't even begin to imagine how we'll be able to run validations on the data.

pstenbjorn:
@jungshadow just to be clear, the current design of the schema doesn't support this, but it it can be modified to permit this usecase.  I also agree with @jktomer and @jdmgoogle that coding to meet specific legislative requirements sounds like far to broad a scope.

jungshadow:
@pstenbjorn Ah...then I retract my "LGTM." Thanks for the clarity.

jktomer:
> @jungshadow just to be clear, the current design of the schema doesn't support this, but it it can be modified to permit this usecase. I also agree with @jktomer and @jdmgoogle that coding to meet specific legislative requirements sounds like far to broad a scope.

> @pstenbjorn Ah...then I retract my "LGTM." Thanks for the clarity.

I think what Paul meant is that the current design doesn't support it, but after this PR it will?

pstenbjorn:
@jktomer yes, exactly.  So it may not LG at the moment, but perhaps in the next day or so.

jdmgoogle:
@jungshadow @pstenbjorn I'm slightly confused as to what's currently missing from the schema that prevents us from supporting CA, OH, or other-non-AZ-style rotations. A `Precinct` will link to a `BallotStyle` which has a set of `OrderedContest` entries; these `OrderedContest` entries should be placed in the XML in the order in which those `ContestType`s appear in that particular `BallotStyle`. Within each `OrderedContest` the set of `CandidateSelection`/`BallotMeasureSelection`/`PartySelection` elements will be the order in which the `BallotSelectionId`s are referenced.

E.g., (with great handwaving) assuming
Candidate "Alice" (can1)
Candidate "Bob" (can2)
Candidate "Carol" (can3)
BallotMeasureSelection "Yes" (bms1)
BallotMeasureSelection "No" (bms2)
CandidateContest "Dogcatcher" (cc1)
BallotMeasureContest "Free Puppies For All?" (bmc1) *[secretly funded by incumbent Alice to drive up dogcatcher business and give her photo ops]*
OrderedContest "Dogcatcher" (oc1) ["cc1", ["can1", "can2", "can3"]]
OrderedContest "Dogcatcher" (oc2) ["cc1", ["can2", "can3", "can1"]]
OrderedContest "Free Puppies" (oc3) ["bmc1", ["bms1", "bms2"]]
BallotStyle "bs1" ["oc1", "oc3"]
BallotStyle "bs2" ["oc2", "oc3"]

The two ballot styles reference the same contests in the same order, but the first ballot style lists the candidates in the order "Alice", "Bob", "Carol" while the second ballot style lists the candidates in the order "Bob", "Carol", "Alice".

This is my understanding of the current layout. Is this understanding (a) incorrect, or (b) failing to capture the CA/OH/non-AZ rotations accurately?

pstenbjorn:
@jdmgoogle sorry, I think you have captured it in this PR.  I think I need to create a sample XML to see prove the concept.  But the logic you outlined seems to capture the needed usecase.

cjerdonek:
@jdmgoogle That's my understanding, too.  I think @pstenbjorn was only saying that the spec didn't support jurisdictions that vary the order _within_ a precinct (which is apparently what Arizona does?).

jungshadow:
@cjerdonek I think we're on the same page. This PR schema change does handle rotation at the precinct-level, but not the AZ-model. That was my initial interpretation and I was confused by the conversation.

@pstenbjorn You're right. We need a reference implementation as this issue seems to be near-and-dear to everyone's heart. You may be the one person here with easy access to real state data. Any chance you might be able to put that together (assuming @jdmgoogle doesn't mind another committer to the branch)?

jdmgoogle:
@jungshadow In general I don't mind another committer to the branch, but we still need to transform the sample_feed.xml file to use the 5.0 spec. I know @pstenbjorn had a pass at making the spec use a Venetian Blind layout (per #120).

Here's my proposal:

1) Assuming that the one-line change in this PR (http://38.media.tumblr.com/tumblr_m5gu9pq9qg1qh71k6o1_500.gif) is okay, let's commit it.

In parallel, then

2) I'll update the sample_feed.xml to validate using the post-this-PR schema; and
3) @pstenbjorn can update #120 to be based off the post-this-PR schema.

Depending on which one of us finishes first, either (a) I'll merge in @pstenbjorn's changes and update the sample_feed.xml to validate, OR (b) @pstenbjorn can merge in my changes and update the sample_feed.xml to meet the post-PR #120 schema.

jungshadow:
[Sounds like a plan.](http://bukk.it/reagan-approves.jpg)

pstenbjorn:
Sounds good.  @jdmgoogle I will be happy to generate post #120 update

jdmgoogle:
Okay. Committing this and doing a cleanup of sample_feed.xml.

pstenbjorn:
I will pull the .xsd and redo the #120 PR based on the latest version with style changes as mentioned elsewhere.

Issue #45, opened 2015-03-12
Create a "best practices" guide for creating a VIP feed.


jungshadow:
This is something that @jktomer [brought up](https://github.com/votinginfoproject/vip-specification/issues/44#issuecomment-78379427) in #44. If we move to `xs:ID` and `xs:IDREF`, this could be helpful in having more visually-useful prefixes other than an underscore or a random letter.

jdmgoogle:
This seems to be a "best practices" for creating a VIP feed rather than having to do with the names of the XML objects, elements, or attributes. If that's the case we should have a bug for "create a 'best practices' guide."

jungshadow:
@jdmgoogle Absolutely. Just wanted to have a placeholder to remind me.

jdmgoogle:
Note that @pstenbjorn started this with #225.

jdmgoogle:
@pstenbjorn In your opinion, what is left to do on this issue?

pstenbjorn:
@jdmgoogle I think this should be closed.

jdmgoogle:
@jungshadow @jen-tolentino @cjerdonek, do you have any additional comments?

jungshadow:
Let's close. We can open other issues when we need to add something new/specific or if we need to fix/remove something.

PR #130, opened 2015-04-30
Make sample feed validate

jdmgoogle:
The sample feed now validates. My thoughts:

- This isn't the final sample feed because we still need to figure out some best practices for ordering the top-level objects (e.g., Source, ElectionAdministration, Locality, etc, etc) within the file;
- We should probably alphabetize the order of elements as much as possible just to make it easier to remember the order in which elements go;
- Breaking up names into components is *painful* and I expect most users will just end up throwing everything into one of the name buckets (it's what I would do); and
- ContactInformation should probably be inlined instead of an IDREF everywhere.


cjerdonek:
1. Alphabetizing the top-level objects and elements seems fine to me.  Note that alphabetizing the type _definitions_ is already in the style guide (albeit within the groups of enumeration, simpleType, complexType).

2.  What are you referring to when you say break names into components?  Are you talking only about the Person object?

3. I agree with you re: in-lining ContactInformation.  Note that it still might make sense to have an "identifier" attribute on the instances.  This can be used for cross-referencing purposes when flat files are used to specify the instances.  This is also what we do for InternationalizedText.

jdmgoogle:
> Alphabetizing the top-level objects and elements seems fine to me.

I was actually just thinking of alphabetizing the elements within an object. As for the top-level object ordering, I'm up for discussing that.

> What are you referring to when you say break names into components? Are you talking only about the Person object?

Correct. Breaking the `Person` name into parts is time-consuming and is likely to just end up having users cram everything into one of the fields.

> Note that it still might make sense to have an "identifier" attribute on the instances

I agree. I move we keep the identifier but inline it otherwise.

jktomer:
> Breaking up names into components is painful and I expect most users will just end up throwing everything into one of the name buckets (it's what I would do); and

I expect that people making exports from an EMS that has separate name components might use name components, and that everyone else (which might well be "everyone" as I don't know that such EMSs exist) will use the formatted ballot name only.

cjerdonek:
Some comments on the PR (it seems like the file is too long for GitHub to allow in-line commenting):

Hyphens in the zip codes:

    <AddressLine>Charlottesville, VA 229118434</AddressLine>

Why left-pad the numbers?

    <Precinct id="pre90139">
        <Name>103 - BRANCHLANDS</Name>
        <Number>0103</Number>

The top-level objects of a given type aren't consecutive in the sample (e.g. CandidateSelection instances are scattered throughout).  Should the spec be allowing that?

cjerdonek:
Also, I vote for 2-space indents on XML files as well (the style guide currently only has this for the XSD).

cjerdonek:
<HoursOpen id="sched0000">

Should the prefix be something like "hours" or "ho"?

jdmgoogle:
> Hyphens in the zip codes:
> Why left-pad the numbers?

The actual contents of the fields are unchanged from the `vip5` branch:

https://raw.githubusercontent.com/votinginfoproject/vip-specification/vip5/sample_feed.xml

(Minus breaking names up, of course.)

> The top-level objects of a given type aren't consecutive in the sample (e.g. CandidateSelection instances are scattered throughout). Should the spec be allowing that?

I'm agnostic on that. I can see requiring that we bunch instances together, but at the same time it may be useful to allow feed creators to put office data, then contest data, then candidate data grouped together (for example).

> Also, I vote for 2-space indents on XML files as well (the style guide currently only has this for the XSD).

I'm fine with that.

> Should the prefix be something like "hours" or "ho"?

hrs?

cjerdonek:
> hrs?

I don't think we have a guideline on this yet, so it might be worth hashing it out now (perhaps opening an issue just for the prefixing convention).  Personally, I think it'd be nice if the convention could be algorithmic to the extent possible so that each one doesn't need to worked out individually (e.g. leading characters, or initials for multi-word names).

jdmgoogle:
I see prefixes as being part of "best practices" rather than being a hard-and-fast rule (i.e., this is trying to solve part of #45).

cjerdonek:
Yes, that is what I was referring to -- hashing out that one aspect of the best practices.

jdmgoogle:
The history of this PR and branch has gotten to be a mess because github sucks at rebasing XML. I'm nuking this PR and branch and starting over again.

cjerdonek:
FYI, you can always force push to a PR branch if you want to start over: `git push -f`

PR #131, opened 2015-04-30
Venetian Blind conversion, remove id, rename OfficeBase

pstenbjorn:
This is are replacement of PR #120 to include history

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -13,13 +13,13 @@ reference elements for polling location
         <xs:element name="Source" minOccurs="1" maxOccurs="1">
           <xs:complexType>
             <xs:all>
+              <xs:element name="Name" type="xs:string" />
+              <xs:element name="VipId" type="xs:string" />
               <xs:element name="DateTime" type="xs:dateTime" />
               <xs:element name="Description" type="InternationalizedText" minOccurs="0" />
-              <xs:element name="FeedContactInformation" type="ContactInformationType" minOccurs="0" />
-              <xs:element name="Name" type="xs:string" />
               <xs:element name="OrganizationUrl" type="xs:anyURI" minOccurs="0" />
+              <xs:element name="FeedContactInformationId" type="xs:IDREF" minOccurs="0" />
               <xs:element name="TermsOfUseUrl" type="xs:anyURI" minOccurs="0" />
-              <xs:element name="VipId" type="xs:string" />
             </xs:all>
             <xs:attribute name="id" type="xs:ID" use="required" />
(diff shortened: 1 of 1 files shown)


cjerdonek:
This PR still seems like it is trying to address multiple issues at once (e.g. alphabetizing element names which isn't part of our style guide yet, changing type names, as well as addressing parts of our Venetian blind style but not all).  Because of this, it makes the diff bigger and harder to evaluate and see what is changing and why (as well as what changes aren't being included and why).

Regarding Venetian blind, our style guide says for the type definitions to go at the beginning before the root element, but this PR has them going after the root element.  This is a comment I made on the previous PR as point (4) in [this comment](https://github.com/votinginfoproject/vip-specification/pull/120#issuecomment-94949170).  The style guide also says that the type definitions should occur alphabetically within each category, something I said as point (3) in the same comment.  It also seems like not all types have been converted (e.g. State).

cjerdonek:
The PR also has things like "CandidateSelection" changing to "CandidateSelectionType" and "BallotStyle" changing to "BallotStyleBase," which I don't think are the correct changes.  For reasons I mentioned previously, I would encourage the type name changes to be done as a separate PR after the Venetian Blind changes since the type name changes are about a different guideline.

pstenbjorn:
@cjerdonek great, thanks.  @jungshadow and @jdmgoogle please let me know if there are any other issues you see with this PR and then I can open another to address naming standards.

jdmgoogle:
@pstenbjorn I'd like to second @cjerdonek's comments here, in that I'm trying to figure out exactly what this PR does. It seems to do a few things, including

1. Alphabetize the attributes and elements within an object;
2. Rename most objects `Foo` to `FooBase` (which seems to be opposite the desired plan); and
3. Inline most of the objects (including `BallotStyle`s, which also seems to be opposite the desired plan).

I second @cjerdonek's suggestion of breaking this PR into multiple parts. At the very least having each change be a different commit to the same PR would allow us to visualize each type of change (e.g., renaming, reordering, restructuring).

pstenbjorn:
@jdmgoogle @cjerdonek I have rebased the file and will just do the Venetian Blind conversion.

pstenbjorn:
@cjerdonek and @jdmgoogle I am sorry about my confusion.  I have recommitted where the only changes are to format the existing schema as a Venetian Blind schema.  This is based on a rebase of the vip5 branch as of yesterday.

jdmgoogle:
LGTM

jungshadow:
LGTM

PR #133, opened 2015-05-01
Minor changes to the schema to get it to validate with a sample feed.

jdmgoogle:
Some of the elements and attributes needed to have some minOccurs/maxOccurs values tweaked.
Also made the FilingDeadline a date instead of a datetime since that's what was in the previous sample feed.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -137,7 +137,7 @@ reference elements for polling location
       <xs:element name="Ward" type="xs:string" minOccurs="0" maxOccurs="1" />
       <xs:element name="PollingLocationId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
       <xs:element name="IsMailOnly" type="xs:boolean" minOccurs="0" maxOccurs="1" />
-      <xs:element name="BallotStyleId" type="xs:IDREF" minOccurs="1" maxOccurs="1" />
+      <xs:element name="BallotStyleId" type="xs:IDREF" minOccurs="0" maxOccurs="1" />
       <xs:element name="ExternalIdentifiers" type="ExternalIdentifiers" minOccurs="0" />
     </xs:sequence>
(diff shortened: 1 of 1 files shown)


vip_spec.xsd
```diff
@@ -259,7 +259,7 @@ reference elements for polling location
       <xs:element name="PostElectionStatus" type="CandidatePostElectionStatus" minOccurs="0" />
       <xs:element name="PreElectionStatus" type="CandidatePreElectionStatus" minOccurs="0" />
       <!-- NOTE: Still trying to get feedback or an example of what this means and how it's used -->
-      <xs:element name="SequenceOrder" type="xs:integer" />
+      <xs:element name="SequenceOrder" type="xs:integer" minOccurs="0" />
```
jktomer: As mentioned in discussion on the PR that introduced ballot style IDs to precinct, I think we can safely scratch this altogether, but that's probably a topic for a separate PR.

vip_spec.xsd
```diff
@@ -421,7 +421,7 @@ reference elements for polling location
       <xs:element name="OtherVoteVariationType" type="xs:string" minOccurs="0" />
       <xs:element name="ElectorateSpecification" type="InternationalizedText" minOccurs="0" />
       <!-- NOTE: Still trying to get feedback or an example of what this means and how it's used -->
-      <xs:element name="SequenceOrder" type="xs:integer" />
+      <xs:element name="SequenceOrder" type="xs:integer" minOccurs="0" />
```
jktomer: ditto above.

jktomer:
LGTM

vip_spec.xsd
jdmgoogle: Yeah, basically I'm doing the minimum amount of work just to get the schema to a point where a feasible sample feed can validate.

jungshadow:
LGTM

Issue #60, opened 2015-03-20
Create separate CSV documentation


jungshadow:
Does the CSV format deserve its own documentation or should it be integrated with the XML spec documentation? I can see a lot of overlap, so it may make sense to keep them integrated, but the new spec may also make the resultant CSVs an animal in and of themselves.

cjerdonek:
Personally, I think the CSV format could use its own page, even if the page starts out only as an overview with high-level guidance.

Given that `index.html` is quite large, one thing I was wondering about if there is some way it can be broken up and made DRY and easier to maintain.  For example, it could be made into a template with the chunks for each object type plugged in (a bit like the CSV and XML data sources, an approach I really like by the way).  Moving in this direction would open up the possibility for having different "views" of the spec (e.g. CSV and XML views), so that the same information wouldn't need to be maintained in two different places.

But for starters, I think a "CSV Overview" page would be good and easier to do.

jdmgoogle:
Temporarily assigning @nomadaisy to this as she seems to be most familiar with the CSV format.

Noma: feel free to assign this to someone else if either someone else would be more knowledgeable about it or you're currently overloaded.

jungshadow:
@nomadaisy Let me know if you want to talk through the documentation structure. If you need a Sphinx/rst primer, I can help there as well.

nomadaisy:
@jungshadow That would be great. Let's schedule something.

afsmythe:
This can be closed. There was already separate CSV documentation, and with PR https://github.com/votinginfoproject/vip-specification/pull/376 the CSV spec has been codified within the YAML files.

PR #135, opened 2015-05-01
Create initial doc structure

jungshadow:
Adding initial doc tree and included generated files for ease of publishing. No real content, yet. Had to add a line to .gitattributes to deal with Windows line endings in `make.bat`. Semi-related to #60.


jungshadow:
I wouldn't spend too much time looking this over. It's just the structure and some generated files.

jdmgoogle:
Oh, let me just peek at some of the CSS/JS ...



LGTM

jen-tolentino:
LGTM!

jungshadow:
@jdmgoogle :100:

cjerdonek:
FYI, I've used Read the Docs before.  I don't think you're supposed to check in the build folder (i.e. `docs/_build`), only the source files.

jungshadow:
@cjerdonek Wouldn't normally be my preference, either. Since we're unable to use RtD for now, we may just want to just copy the generated files to gh-pages (also not my preference). Having the generated files in the repos will make that process easier for more folks, rather than have to set up sphinx and run the build. We'll kill them off once we're cleared (assuming that happens).

cjerdonek:
Yes, committing the build files to gh-pages would be much better so as not to clutter the history, etc, if that's what you're suggesting.

cjerdonek:
In the past, one workflow I have used is to have two copies of a repo on my machine in sibling folders: one permanently set to the gh-pages branch, and one set to a normal branch.  Then, when working in the normal repo, you can easily set sphinx to output to the build folder in the other repo, e.g. `../vip-gh-pages/docs/_build` or whatever.

Issue #136, opened 2015-05-01
Order type definitions in XSD


cjerdonek:
We need to order the type definitions in the XSD per the style guide.  Is it okay for me to do this, or is someone else already working on a big change to the XSD?

jdmgoogle:
Not sure what the status is of @pstenbjorn's Venetian-Blind-izing the XSD.

pstenbjorn:
@cjerdonek @jdmgoogle Venetian Blinding is merged.  I am fine with @cjerdonek ordering the XSD.

cjerdonek:
I will create a PR now.

cjerdonek:
Started PR #137.

jdmgoogle:
Since this should be a relatively non-controversial set of PRs (i.e., just sorting some elements and attributes), would @jungshadow or @pstenbjorn object if we waived the normal two-reviewer rule and I reviewed @cjerdonek's PRs by myself?

jungshadow:
Fine by me.

pstenbjorn:
Same here

cjerdonek:
Closed by PR #137.

PR #137, opened 2015-05-01
Address issue #136: order type definitions

cjerdonek:
Address issue #136.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -5,6 +5,120 @@ added ocdid to district
 reference elements for polling location
 -->
 <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" version="5.0.1" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+  <!-- Start enumeration types. -->
+  <xs:simpleType name="BallotMeasureType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="ballot-measure" />
+      <xs:enumeration value="initiative" />
+      <xs:enumeration value="referendum" />
+      <xs:enumeration value="retention" />
+      <xs:enumeration value="other" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="CandidatePostElectionStatus">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="advanced-to-runoff" />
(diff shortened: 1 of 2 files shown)


cjerdonek:
Will do the other types next.  Just wanted to start this.

vip_spec.xsd
```diff
@@ -5,6 +5,120 @@ added ocdid to district
...
+      <xs:enumeration value="utility" />
+      <xs:enumeration value="ward" />
+      <xs:enumeration value="water" />
+      <xs:enumeration value="other" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="IdentifierType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="ocd-id" />
```
jdmgoogle: Are you ordering the enums only by the name of the enum, or by the name of the enum and then the values of the enum (with 'other' always last)?

vip_spec.xsd
```diff
@@ -5,6 +5,120 @@ added ocdid to district
...
+      <xs:enumeration value="fips" />
+      <xs:enumeration value="national-level" />
+      <xs:enumeration value="state-level" />
+      <xs:enumeration value="local-level" />
+      <xs:enumeration value="other" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="OebEnum">
```
jungshadow: Basing this on the styleguide enumeration ordering, should this be ["both", "even", "odd"]?

cjerdonek:
I believe this is done now (just ordering the types).  Name changes can occur in subsequent PRs.

vip_spec.xsd
cjerdonek: I am only ordering the "chunks" in this PR and not changing the internal ordering of any type or enum definitions (which can happen in subsequent PR's if needed).

vip_spec.xsd
cjerdonek: Sorry if repeating, but yes, this can happen subsequently.  In this PR, I am restricting only to moving the chunks in this PR as is -- and leaving changes within type definitions to subsequent PR's.  This will make it easier to see and review changes within type definitions, as opposed to having moves combined with changes in the same PR.

jdmgoogle:
LGTM

Issue #142, opened 2015-05-01
Address remaining local and global types


cjerdonek:
I noticed that OrderedContest should be in-lined inside the BallotStyle definition.  Also, State needs to be made a named type rather than anonymous.

jdmgoogle:
I'm not sure that OrderedContest needs to be inlined.
The elements within each object need to be sorted.
Also ContactInformation should be inlined.
I'm running out of steam right now, though, so let's pick up this last cleanup on Monday.

cjerdonek:
Also, the Term object should be in-lined inside Office.

cjerdonek:
Yeah, I'm just making a note for Monday, etc.  OrderedContest is used only within BallotStyle and doesn't have an ID attribute (and doesn't fit any of the criteria in the style guide), which is why it should be in-lined.

cjerdonek:
On second thought, I will start a PR for this.

cjerdonek:
Created PR #143.

Regarding ContactInformation, I checked, and I don't think it can/should be in-lined because it is used/referenced in more than one type: Person, Source, Department, Department.VoterService, and Office.

PR #143, opened 2015-05-01
Address issue #142: fix some local / global type definitions.

cjerdonek:
Address issue #142.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -160,7 +160,14 @@ reference elements for polling location
 
   <xs:complexType name="BallotStyle">
     <xs:sequence>
-      <xs:element name="OrderedContest" type="OrderedContest" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="OrderedContest" minOccurs="0" maxOccurs="unbounded">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="ContestId" type="xs:IDREF" />
+            <xs:element name="BallotSelectionId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
       <xs:element name="PartyId" type="xs:IDREF" minOccurs="0" maxOccurs="unbounded" />
       <xs:element name="ImageUri" type="xs:anyURI" minOccurs="0" />
     </xs:sequence>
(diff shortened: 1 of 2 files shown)


vip_spec.xsd
```diff
@@ -160,7 +160,14 @@ reference elements for polling location
 
   <xs:complexType name="BallotStyle">
     <xs:sequence>
-      <xs:element name="OrderedContest" type="OrderedContest" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="OrderedContest" minOccurs="0" maxOccurs="unbounded">
```
jdmgoogle: OrderedContest should not be inlined, it should have an ID added to it. Otherwise with **N** different ballot styles, you'll be repeating the Presidential, Gubernatorial, Attorney General, etc, races **N** times. OrderedContest should be a stand-alone item with an ID and should be referred to by the ID.

I'm not sure how this slipped through in previous iterations, but reference-by-ID is the right way to go.

jdmgoogle:
Please undo this PR. This is headed in the opposite direction that we're going with the structure and style of the standard, and in some instances (E.g., inlining `OrderedContest`) is actually significantly less-than-optimal.

vip_spec.xsd
cjerdonek: K, I'm awake, so I'll fix this now.

jdmgoogle: And I thought *I* woke up early. :)

cjerdonek:
Updated.

cjerdonek:
Oh, and I can also go ahead and in-line ContactInformation (i.e. remove the ID / IDREF).  I was conflating in-lining the type definition (which is what I was saying shouldn't be done) with in-lining the data.

jdmgoogle:
Thanks. Could you also update `BallotStyle` to have an `OrderedContestId` element, and a top-level element to list all the `OrderedContest` elements?

Once you've done that I can create a PR to alphabetize the elements within each object and update the sample schema.

cjerdonek:
Sure.

jdmgoogle:
WRT `ContactInformation` I mean to define it in the schema outside of any particular element.

I'm not doing a good job of explaining what I'm trying to do, so just leave it alone in this PR and I'll create my own PR to make the change.

cjerdonek:
> WRT `ContactInformation` I mean to define it in the schema outside of any particular element.

It already is.  I assumed you meant to handle it how "InternationalizedText" is handled, where there is no top-level element listing all objects.

cjerdonek:
Updated.

jdmgoogle:
LGTM. Paging @jungshadow @pstenbjorn for review.

cjerdonek:
> Once you've done that I can create a PR to alphabetize the elements within each object and update the sample schema.

By the way, in the same way we have "other" at the end as an exception to the alphabetization rule, it would be nice if elements named "OtherFoo" can always follow "Foo" in the case of enumerations using an other.

cjerdonek:
@jdmgoogle can you merge this on your own when you get approval?  I might not be around when you hear back.

jdmgoogle:
Will do.

pstenbjorn:
@jdmgoogle LGTM

PR #144, opened 2015-05-04
Un-Venetian-Blind ContactInformation

jdmgoogle:
It makes more sense to have the content included with the elements instead of mentioned by reference.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -224,7 +224,7 @@ reference elements for polling location
       <xs:element name="Uri" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
       <xs:element name="Email" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
     </xs:sequence>
-    <xs:attribute name="identifier" type="xs:ID" use="required" />
+    <xs:attribute name="identifier" type="xs:string" use="required" />
   </xs:complexType>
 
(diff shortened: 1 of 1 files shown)


cjerdonek:
This isn't actually "un-Venetian-Blinding" it since the type definition is at the top level.

My only suggestion is to change the "id" attribute on the ContactInformation object to the following, as we discussed previously:

    <xs:attribute name="identifier" type="xs:string" />

This is because the ID is no longer a required ID, but can be present to link the data to a flat file, if a flat file was used.  This is the same pattern that InternationalizedText uses.

cjerdonek:
After that, LGTM.

pstenbjorn:
@jdmgoogle I agree with @cjerdonek assessment but my question is what is the value of including an identifier if these elements are not referenced by ID?

cjerdonek:
If the objects are provided as flat files, providing an identifier lets you link to the data back to the row of the flat file it came from.  Otherwise, there is no definitive way to link it back.

jdmgoogle:
Good idea. Done.

jdmgoogle:
Whoops. Good catch (even though xs:ID is just a subset of xs:string).

cjerdonek:
Thanks, LGTM.

jdmgoogle:
Based on the comments from @cjerdonek and @pstenbjorn I'm assuming this meeting PR review guidelines. Merging.

PR #145, opened 2015-05-04
Alphabetize the elements in the spec

jdmgoogle:
Alphabetize the elements in the spec (with two small changes).

Updated the style guide to reflect this ordering.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -113,8 +113,8 @@ For example--
 Within a type, elements should be listed first, followed by attributes. Each
 should be in alphabetical order, with only two exceptions:
 
-1. If an element named `Type` or `FooType` is an enumeration which contains a value
-   `other`, then `OtherType` or `OtherFooType` should be placed after `Type`.
+1. If an element named `Foo` is an enumeration which contains a value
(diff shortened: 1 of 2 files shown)


pstenbjorn:
LGTM

STYLEGUIDE.md
```diff
@@ -111,14 +111,25 @@ For example--
 ### Element/Attribute Ordering
 
 Within a type, elements should be listed first, followed by attributes. Each
-should be in alphabetical order:
+should be in alphabetical order, with only two exceptions:
+
+1. If an element named `Type` or `FooType` is an enumeration which contains a value
+   `other`, then `OtherType` or `OtherFooType` should be placed after `Type`.
```
cjerdonek: I would simplify this to say if an element named `Foo` ... then `OtherFoo` ....  The reason is that doesn't really matter if the type name ends in "Type" (and it needn't end that way).

cjerdonek:
(Sorry for the duplicate comment, it disappeared on GitHub then came back.)  Aside from that one comment, LGTM.

jdmgoogle:
@cjerdonek Updated.

Merging now.

PR #146, opened 2015-05-05
Updated sample feed

jdmgoogle:
Updated the sample feed to contain pretty much exactly the same data, just in VIP 5.0 format. I'm trying to figure out some of the best practices in terms of ID prefixes and ordering (see #45), but this is me planting the flag in the ground.


jungshadow:
This looks good, @jdmgoogle. My only comments are on style:

* It's so nice to be able to browse the file in GitHub :100:
* :+1: Hungarian notation
* In keeping with the styleguide, should the indentation be 2 spaces?

cjerdonek:
Looks okay to me.  I'd prefer the top-level elements to be alphabetized, but that's a cosmetic thing.

cjerdonek:
Or grouped at least.

jdmgoogle:
@jungshadow I can re-indent.

@cjerdonek I tried to group in a way that would make the most sense. I ended up going with
=== Administrative
Source
Election
State
Locality
ElectionAdministration (and Person associated with that)
HoursOpen
=== Ballot-related items
Party
ElectoralDistrict
Office
Contests
BallotStyle
OrderedContests
(CandidateSelection, Candidate, Person) tuples
=== Voting-location-related items
PollingLocation
Precinct

cjerdonek:
By "grouping" I meant all top-level objects of one type appearing alongside the others of that type in the same group, e.g. Person objects appearing within a single group next to each other, as opposed to being scattered throughout.

In 1622.2, this happens automatically for them because their XSD has a top-level PersonCollection object, which ensures that all Persons are grouped.

While it may be aesthetically nice to group Persons alongside the objects they relate to, this may be impractical to recommend for real-world feeds.

jdmgoogle:
Why would it be impractical?

cjerdonek:
Because (1) it would need to be documented separately, (2) it wouldn't be enforceable via the spec, (3) there are edge cases that wouldn't be workable (e.g. a Person object referenced in both categories, etc), (4) it would require custom code to ensure that feeds are outputted in that order, and (5) it may not be useful in practice anyways since feeds aren't designed for human-readability (e.g. the files can be huge).

jdmgoogle:
I'm not sure how we got from "I'm ordering the elements in this one example feed in a way that will help people understand how the elements in this new spec relate to each other" to "we **must** ensure that all feeds produced by end-users are in this **exact same ordering***."

cjerdonek:
I know earlier you had mentioned wanting to document best practices for feeds, so my points were speaking to whether this should be included as a best practice.

Regarding this particular sample feed, I think there's value in having the sample look and be ordered like real-world feeds, otherwise people might get the impression that real-world feeds need to ordered this way, too.  If you want to people the sample feed in a different order, I think it would help to include a comment at the top to the effect of, "elements have been ordered to aid human understanding of the sample."

jdmgoogle:
I suspect there's going to be a variety of real-world feed organizations, and people will organize them in whatever way makes sense to them. The code that parses them (at least on Google's end) will be agnostic to how the elements are ordered, provided that the feed validates.

Ultimately I suspect that people will primarily use the sample feed to understand how the elements of VIP 5.0 relate to each other. I agree it's likely that initial adopters will model their first feeds on the sample one we provide here, but once they've gained experience with the new format and automated the process to generate feeds, their feeds will diverge in look and feel from the sample one, and that's okay.

cjerdonek:
And how about with regard to best practices?  In the first comment to this PR, you wrote--

> I'm trying to figure out some of the best practices in terms of ID prefixes and ordering

That's why I was raised those points as to why I didn't think we should recommend such an ordering as a best practice.

jdmgoogle:
To clarify, there are some separate issues here:

1. How do we help people who are new to VIP 5.0 understand the format and how the elements relate to each other?
2. What ordering of elements are reasonable in helping people understand the feed that has been generated (i.e., is it possible to formulate a feed that is not completely hostile to manual debugging)?
3. What are best practices in creating a feed for real-world use?

I see the sample feed as mainly addressing (1) with a bit of (2) and (3). There are aspects of (3) that are present in the sample feed, such as Hungarian Notation in the IDs. There are aspects of (2) in the sample feed as well, in that grouping [CandidateSelection, Candidate, Person] elements together makes it possible to ensure that a CandidateSelection references the correct Candidate which in turn references the correct Person.

Early in the life of the standard I believe it'll be beneficial -- but not necessary -- to encourage -- but not require -- that feeds be structured in a way that is not hostile to manual debugging; some of the sample 1622.2 feeds were structured this way and it was painful to dig through them. As time goes on I suspect and hope that more automated tools will come into existence which allow for complex troubleshooting of non-intuitively structured feeds.

In the short term, though, I believe that creating sample feeds that have fully stratified element layout will (a) make it more difficult for people to understand the feed format -- i.e., failing at goal (1) -- (b) subtly encourage people to create their own feeds which are similarly difficult to understand manually -- i.e., failing at goal (2) -- and provide no explicit benefit in creating feeds which can be parsed by computers.

cjerdonek:
I agree that a sample feed can serve different purposes (e.g. pedagogical versus showing an example of something found "in the wild").  Indeed, some of these purposes can even be at odds.  And I agree that mainly focusing on (1) is worthwhile in this case.  It's just that, currently, which of these purposes the feed is fulfilling isn't self-evident when looking at the feed alone.

Thus, I'm advocating including a note like the following in the comments at the beginning of the feed, "This feed has been ordered to aid the reader in understanding the format and the relationship of the elements to one another."  Without a note like that, it's not clear if the ordering of the sample feed is pedagogical versus being recommended as a best practice (which I think we should discuss before implying that it is a best practice).

jdmgoogle:
Understood. I've updated the header comment to reflect this decision.

cjerdonek:
Thanks, LGTM.

jdmgoogle:
With the LGTM from @jungshadow and @cjerdonek, I'm merging this.

Issue #149, opened 2015-05-05
style guide: document "identifier" pattern for in-lined objects


cjerdonek:
I think it would be worth adding a note about our pattern of adding an optional "identifier" attribute with type string for objects whose data is in-lined and that do not have an ID.  When flat files are used, the identifier can be used in lieu of an ID to link the data back to the row of the flat file it came from.

jdmgoogle:
Go for it. All yours. :)

cjerdonek:
Will do, maybe not today but within the next few days. :)

cjerdonek:
Closed by PR #229.

PR #150, opened 2015-05-05
Rename all FooUrl to FooUri

jdmgoogle:
Doing this for (a) consistency, and (b) 1622.2 convergence. (Also so @pstenbjorn won't be cursing at us six months from now.)


cjerdonek:
LGTM.  I can also add a note to the style guide re: the preference for Uri vs Url as a rider when I make the PR for issue #149.

jungshadow:
LGTM

PR #152, opened 2015-05-06
Pull HtmlColorString into its own type for sake of easier documentation.

jdmgoogle:
See $SUBJECT.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -118,6 +118,12 @@
   </xs:simpleType>
 
   <!-- Non-enumeration simple types. -->
+  <xs:simpleType name="HtmlColorString">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="[0-9a-f]{6}" />
+    </xs:restriction>
+  </xs:simpleType>
+
   <xs:simpleType name="TimeWithZone">
     <xs:restriction base="xs:string">
       <xs:pattern value="(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]|(24:00:00))(Z|[+-]((0[0-9]|1[0-3]):[0-5][0-9]|14:00))" />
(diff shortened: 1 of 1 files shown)


cjerdonek:
FYI, according to the style guide this type definition should be kept in-lined, since it's only used in one place.  Re: documentation, it can simply be documented at the one place it's used (when documenting that element of Party).

jdmgoogle:
By "easier documentation" I mean "easier to write some Python that generates all the RSTs for us" (which I did this morning). And I'm not sure I agree with the decision to in-line definitions like this. We've already pulled the TimeWithZone pattern out on its own; putting that back in for the sake of style guide compliance means that god-awful-ugly pattern gets inlined and used twice, and has to be broken up over two lines just to match the other part of the style guide. So we're already violating the style guide in one way, which means either the XSD needs to change or the style guide needs to change. I vote for the latter.

cjerdonek:
TimeWithZone is used in two different places, so it should be broken out.

There's no reason to break this out (for style guide reasons or ease of documentation).  To document Party.Color, you can simply write something like, "the color of the party as an HTML color string," and provide as much detail as you want.

jdmgoogle:
No, TimeWithZone is only used in one place: Hours ~> Schedule ~> OpenHours.

And HtmlColorString as a type is much more descriptive and less likely to cause confusion than "this is an xs:string with some arbitrary pattern enforcement, please go look at the documentation."

cjerdonek:
> No, TimeWithZone is only used in one place:

TimeWithZone is used in two places and exactly matches the example provided in the style guide for InternationalizedText.

From the style guide:

> 2. elements of the type occur in more than one location within the
     root element.  For example--

```xml
<xs:element name="Title" type="InternationalizedText" />
<xs:element name="Subtitle" type="InternationalizedText" />
```

And for TimeWithZone:

```xml
<xs:element name="StartTime" type="TimeWithZone" maxOccurs="1" />
<xs:element name="EndTime" type="TimeWithZone" maxOccurs="1" />
```

"Two places" doesn't mean "in two different elements," for the reason you mentioned above (not having to repeat the same definition in two spots).  It means, two "occurrences."

jungshadow:
I can see this being used in more than one spot (even if it isn't now). I think we might want to amend the styleguide to cover these cases.

cjerdonek:
What's wrong with following the style guide and breaking it out once we _actually_ start using it in two spots?  The Party.Color documentation can be something as simple as, "the color of the party as a 6-digit HTML color code, for example `ffff00` for yellow."

cjerdonek:
Any in-lined type could _potentially_ be used in two spots.  One of the points in our original discussion of the guideline was to keep it "rule-based," so we wouldn't have to have subjective discussions about each type.

jdmgoogle:
See PR #153 to make the style guide treat the different restricted string types the same.

jungshadow:
@cjerdonek I feel that this is one of those "perfect being the enemy of the good" conversations. The section you mentioned in the styleguide starts with a description of "...in the spirit of Venetian Blind..." which says to me "we're not following it to the law and the letter." Similarly, I think this change is in the spirit of the styleguide.

**NB:** As I was finishing this up, @jdmgoogle just created a PR to address this...so maybe we can say it follows it "to the law and the letter," now :)

jktomer:
LGTM

jungshadow:
Still LGTM

PR #153, opened 2015-05-06
Make style guide consistent

jdmgoogle:
Make it consistent with regard to enums and other restricted strings.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -214,7 +214,7 @@ if any of the following are true:
    <xs:element name="Subtitle" type="InternationalizedText" />
    ```
 
-3. the type is an "enum".  For example--
+3. the type is an "enum" or other pattern-restricted string.  For example--
 
    ```xml
    <xs:simpleType name="OebEnum">
@@ -222,6 +222,12 @@ if any of the following are true:
        <xs:enumeration value="odd" />
        <xs:enumeration value="even" />
        <xs:enumeration value="both" />
+     </xs:restriction>
+   </xs:simpleType>
+
+   <xs:simpleType name="HtmlColorString">
+     <xs:restriction base="xs:string">
+       <xs:pattern value="[0-9a-f]{6}" />
      </xs:restriction>
    </xs:simpleType>
    ```


jungshadow:
I'm good with this.

cjerdonek:
What is the benefit of this change?

jdmgoogle:
Consistency of handling restricted strings the same (both enumerations and pattern-restricted), cleaner design, and more useful and intuitive type names.

What's the downside?

cjerdonek:
One downside is that readers of the XSD have to jump around more than necessary.  Instead of the definition being at the one place it is being used, they have to go to a different section.  Regarding consistency, I don't think "only restricted strings" is any more consistent than "only enumerations."

cjerdonek:
> cleaner design, and more useful and intuitive type names.

Also, these argument apply to all types, not just restricted strings.  So if we placed value in these arguments, we should be going to pure Venetian blind, which is completely stratified with every type defined at the top level.  I thought we didn't want that.

jktomer:
LGTM

I'm fine either way, but pattern-restricted strings in particular seem more likely to be reused than many things.

cjerdonek:
> but pattern-restricted strings in particular seem more likely to be reused than many things.

There can easily be pattern-restricted strings that would only make sense to be used in one spot.  It depends on the particular pattern and the particular element, etc.

jdmgoogle:
@cjerdonek Every single enumeration (except DistrictType) is only used in one place. Per your argument

>  Instead of the definition being at the one place it is being used, they have to go to a different section.

we should be inlining every single enumeration except DistrictType.

> I don't think "only restricted strings" is any more consistent than "only enumerations."

I disagree. And at least based on the number of people who have weighed in on this PR, it appears several other people disagree.

> So if we placed value in these arguments, we should be going to pure Venetian blind, which is completely stratified with every type defined at the top level. I thought we didn't want that.

I'm okay with pure Venetian Blind. The "we" are people other than myself -- I'd actually prefer that `Department` and `VoterService` be defined at the top level because as it is I think the XSD is a bit messy, but I was outvoted, IIRC -- but I don't see making the final `simpleType` of `<xs:restriction base="xs:string">` which is not top-level-defined to be top-level-defined as being a sea change shift.

I'll give other people another hour or two to weigh in on this, but right now there are three people in favor of this change and one opposed, as well as having met the PR review guidelines. If anyone else objects and can sway myself, @jungshadow, or @jktomer to not make this change then I'll delete this PR and branch. Otherwise I plan to merge before I leave work this afternoon.

PR #154, opened 2015-05-06
Add generated files

jungshadow:


jungshadow:
This adds @jdmgoogle's generated element files to the documentation branch. Still needs to be curated and arranged. This is off-dev branch, so immediately merging.

Issue #160, opened 2015-05-15
If RetentionContest exists, is BallotMeasureType "retention" necessary?


jungshadow:
I may have missed something in the discussion on this, but we have an element `RetentionContest` and a `BallotMeasureType` value of "retention." If we have the element, do we need the value for a `BallotMeasureContest`? Let me know if I'm crazy here.

jdmgoogle:
I don't believe so. I'm readying a PR to make a few very minor fixes, and this will be one of them.

jdmgoogle:
Fixed by #161.

Issue #155, opened 2015-05-08
InfoUri is required in BallotMeasureContest


pstenbjorn:
Noticed that this element should be optional.

`<xs:element name="InfoUri" type="xs:anyURI" />`

should be 

`<xs:element name="InfoUri" type="xs:anyURI"  minOccurs="0" />`

jdmgoogle:
Fixed by #161.

Issue #118, opened 2015-04-21
Identify and adopt a uniform address standard


jdmgoogle:
Currently we use two main address formats: the simple multi-string "AddressLine" in the Contact object, and a more detailed lines/city/state/zip format in SimpleAddressType. The reason for this is that the Contact object comes from 1622.2, and we (Google) hope to promote 1622.2 as a data standard for international election results reporting, including in jurisdictions where the city/state/zip address format doesn't work. The VIP address elements, though, are used for creating street segments in a very US-centric format; the simplicity of the use case is actually what helped spur adoption, and we don't want to deviate from those roots.

This is a placeholder issue to remind us to locate and evaluate different address standards which could be used in both VIP and the IEEE VSSC efforts.

Cc @johnpwack @kennethmbennett @pstenbjorn since they have previously expressed interest in tackling this (non-trivial) problem.

cjerdonek:
> Currently we use two main address formats: the simple multi-string "AddressLine" in the Contact object, and a more detailed lines/city/state/zip format in SimpleAddressType.

FYI, there are actually three address formats currently being used in VIP:

1. the generic "AddressLines" approach used in the Contact object (from 1622.2, for things like Office addresses),
2. "SimpleAddressType" (used for the address of PollingLocations), which breaks out City / State / Zip, and
3. the even more structured "DetailAddressType" (used in StreetSegment), which also breaks out things like HouseNumberPrefix, HouseNumberSuffix, StreetSuffix, etc.

jdmgoogle:
Circling back to this. To update @cjerdonek's comment from earlier, the current address schemes uses in VIP 5.0 are:

1. the generic `AddressLines` elements in `PollingLocation` and `ContactInformation` (the latter of which is used in the NIST format); and
2. the `StreetSegment` object, which is basically nothing but an address now.

jdmgoogle:
For various reasons progress on 5.1 has slipped (I basically ran out of time in December to work on it). In order to not make big changes leading up to the 2016 election, I'm going to push this back to the 5.2 milestone.

afsmythe:
Pushing this to 5.3 (post 2020 elections) milestone.

afsmythe:
[...]

PR #161, opened 2015-05-15
Small cleanups and a few SURVEY questions

jdmgoogle:
Small changes, including
- Making InfoUri an optional field in BallotMeasure (see #155);
- Making ExternalIdentifier an unbounded repeating element;
- Making values in ExternalIdentifiers required;
- Removing 'retention' as a type for BallotMeasureContest (there's a RetentionContest type for that, see #160);
- Making VotesAllowed optional; and
- Adding two SURVEY questions.


jungshadow:
LGTM

jungshadow:
Oh, to answer the survey, yes, we do need to keep the prefix and suffix because Wisconsin:

    W399S7248 County Road Z, Eagle WI 53119

jdmgoogle:
But how does that work with the (required) StartHouseNumber and EndHouseNumber elements in the parent element?

jdmgoogle:
I just checked and it looks like we (Google) don't use the prefix and suffix in our pipeline, so these fields have never been used as far as I can tell. @jktomer may know more, but it looks like we can remove some more unused fields.

Also I moved isWriteIn to be an element instead of an attribute since it was the only non-identifier, non-actual-XML-attribute (i.e., language) to be an attribute instead of an element. I suspect that it was a 1622.2 conversion issue on my part.

jungshadow:
Never mind, I just realized that we now have UnitNumber, which I think solves this problem. I believe we dealt with these addresses as one-offs (i.e. the numbers didn't fluctuate and start==end). I'd have @pkoms or @nomadaisy weigh in on this as they're more familiar with the current dataset.

Just a note on this, I don't feel the survey discussion holds up the merge, but I would be happy if the surveys moved to their own "Up for Discussion" issue, since I'm not sure how if this PR comment thread is going to be easily found by posterity :) Plus, seems a bit weird to have a future PR request reference another PR request as the basis for a fix (unless a bug was introduced). I don't mind creating the issue, if you need me to.

jdmgoogle:
Filed issue #162 to document the `NonHouseAddress` cleanup. The other cleanups not covered by issues here (Making ExternalIdentifier an unbounded repeating element; Making values in ExternalIdentifiers required; and Making VotesAllowed optional) are trivial fixes to the way things should have been in the first place.

jungshadow:
Sounds good and thanks for filing #162. You still get a "LGTM" from me on this.

jungshadow:
@jdmgoogle If you don't hear anything from @nomadaisy and @pkoms by Monday EOD and you get another LGTM per the contributing process, feel free to merge away.

jktomer:
House number prefix and suffix are a bit of a tricky case. @jdmgoogle is right that we presently ignore them (because our own geo stack doesn't seem to support them), but the standard shouldn't be limited to only the things Google Maps recognizes today.

If there is a useful case to be made for either adding a house number prefix/suffix (open question: how would that interact with nonempty ranges? Presumably the same prefix and suffix gets added to each of them; is that useful in reality? Also note that W399S7248 doesn't have an obvious useful prefix+house number+suffix representation.) or an alternate HouseDesignator string to replace start+end house number (no ranges allowed for non-integral house number).

Regardless, breaking fields out into non-house-address seems pointless; I'm not sure why that's even still a thing, given that DetailAddressType exists nowhere else, so there's no common element to factor out. Anything we care about could live directly in StreetSegment IMO. (I don't insist; what we have is fine.)

Does anyone have a count of addresses covered by existing or past feeds which can't use an integer house number? (If you also have a list, I can see how many of them we can geocode as-is, but again, we shouldn't limit ourselves to being able to represent only data that Google can geocode, since there are many other potential users of this data, and Google's capabilities will change over time too.)

jdmgoogle:
@jktomer It's not just what Google can or can't do right now, and you kind of touched on this in your response:

> (no ranges allowed for non-integral house number)

Right now we primarily use street segments to perform range-based lookups of addresses. Having prefix and suffix elements (beyond the `Unit` element) will limit how we do ranges. E.g., if the range is [1A, 3A], does that mean the addresses are {1A, 2A, 3A}, or is the range {1A, 1B, 1C, ..., 2ZX, 2ZY, 2ZZ, 3A}?

If we do want to represent complicated addresses in 5.0 (as opposed to folding it in to #118 in 5.1) then my only suggestion is to add a `StartHouseNumberPrefix` and `StartHouseNumberSuffix`, and if either of those are present then we ignore `EndHouseNumber`.

jktomer:
Yeah, I think that, like the presence of unit numbers, non-integer house numbers should restrict a segment to single-address only (not entirely rare, as it happens, but obnoxious).

Regardless, this can definitely wait until 5.1 from our perspective.

jdmgoogle:
Pinging @jktomer @cjerdonek @pstenbjorn for review/LGTM.

pstenbjorn:
@jdmgoogle and @jktomer the original commits in this PR LGTM.  I also agree that restructuring additional street segment elements might be better left to 5.1.

jdmgoogle:
Thanks. Merging.

jdmgoogle:
I filed #166 as a reminder to better handle complex and non-numeric addresses in the `StreetSegment` element in 5.1

PR #171, opened 2015-05-22
A first pass at defining the enumerations in VIP 5.

jdmgoogle:
This is a first pass at writing definitions for the enum values. Some of them were difficult because of how simple they seem (e.g., "city" in `DistrictType`) and others were just plain complicated (e.g., `VoteVariationType`).

Pinging @cjerdonek to look over the `VoteVariationType` definitions, which I didn't get all the way through. Some of the values describe *how* people vote (e.g., "cast a ballot for *N* people" or "put a number of points next to each candidate") whereas others describe thresholds for victory once votes have been tallied (e.g., majority, supermajority, plurality). It's a bit of a mess, so any help would be appreciated.


cjerdonek:
@jdmgoogle I would be happy to take a look, but I might not get to it until this weekend.  Regarding VoteVariationType, in the cases where you did not finish, should I wait for you to provide those, or are you asking for suggested language?

jdmgoogle:
@cjerdonek Basically I'm asking for as many comments as anyone is willing to offer on any and all of the `VoteVariationType` definitions, both existing and empty.

johnpwack:
Hi everyone,

I was looking at some of the definitions - Sarah in Wisconsin is wanting to
use 'municipality' as a catch-all for villages and town - maybe the
definition ought to be along the lines of "For general purposes
administrative divisions [jurisdictions] that are not villages, towns, or
townships, or used to generally denote any general purpose division
[jurisdiction]."

Cheers, John

*---*

*John P. Wack*

johnpwack@gmail.com

301.640.6626 - cell

301.975.3411 - office

jdmgoogle:
@johnpwack Does Wisconsin not have separate legal definitions for "villages", "towns", and "municipalities"? Or is this something where the state feels that the marginal benefits of the increased accuracy isn't worth the amount of work that'd have to go into maintaining the label data?

jdmgoogle:
Also, a quick ping to @jktomer @cjerdonek @jungshadow @pstenbjorn @nomadaisy @pkoms and anyone else who'd like to review both `VoteVariationType` and `DistrictType` descriptions.

jungshadow:
@jdmgoogle The `DistrictType` definitions look fine to me (also all the others), but I'll punt to @cjerdonek and maybe @johnpwack (if he wants to weigh in) on the `VoteVariationType`.

cjerdonek:
I'll post a comment later today.  Just haven't gotten around to it yet.

docs/xml/enumerations/district_type.rst
```diff
@@ -9,41 +9,43 @@ DistrictType
 | Name                 | Description                                                                      |
 |                      |                                                                                  |
 +======================+==================================================================================+
-| city                 |                                                                                  |
+| city                 | A city.                                                                          |
 +----------------------+----------------------------------------------------------------------------------+
-| city-council         |                                                                                  |
+| city-council         | A city, town, or village council, either in its entirety or for a specific seat. |
```
jktomer: What does "in its entirety" mean here?

city-council means "a city council district," not "the city" (which would be city). At-large city council members are (by definition) elected citywide, so a contest for an at-large city council seat would be associated with the electoral district for the city, which would have type "city."

cjerdonek:
So I thought about this a bit.

Regarding `VoteVariation`, my advice is to document what the enumerations _stand for_ when possible, as opposed to trying to define the terms that the enumerations stand for.  For example, instead of trying to _define_ approval voting, I would simply say that "approval" stands for approval voting.

Here are proposals for each `VoteVariation` value:

* 1-of-m: A voting system where each voter can select at most one candidate.
* approval: Approval voting.
* cumulative: Cumulative voting.
* majority: A contest where the winning threshold is 50%.
* n-of-m: A voting system where each voter can select at most _n_ candidates, where _n_ is the number of seats to elect.
* plurality: A contest where the candidate or candidates with the most votes wins.
* proportional: A proportional representation system (e.g. party list or the single transferable vote).
* range: Range voting.
* rcv: Ranked choice voting, instant runoff voting, or the single transferable vote.
* rcv-borda: Borda count.
* super-majority: A contest where the winning threshold is some percent greater than 50%.
* other: Other.

I would also clarify that more than one enumeration value can apply to a given contest, for example "1-of-m" and "plurality" (in the case of a first-past-the-post contest), or "proportional" and "rcv" (in the case of an STV election with more than one winner).

jdmgoogle:
@cjerdonek I wasn't necessarily trying to *define* the terms, but I was trying to provide some guidelines to help election officials understand when to use each one. Also:

> n-of-m: A voting system where each voter can select at most n candidates, where n is the number of candidates to elect.

What if a voter can select *n* candidates, but *p* will actually end up winning, where *n* != *p*?

Also, right now we only allow on `VoteVariationType` value per contest. I'm tempted to just mirror the definitions in 1622.2 for the sake of completeness, and put a note in the documentation that these will probably end up getting nuked and replaced with whatever comes out of another one of the 1622 working groups.

docs/xml/enumerations/district_type.rst
jdmgoogle: Fair enough. Thanks.

cjerdonek:
> I wasn't necessarily trying to define the terms, but I was trying to provide some guidelines to help election officials understand when to use each one.

Okay, I was referring to cases where you described some aspect of the system without actually naming the system.  For example--

* approval: A voter can cast votes for any number of candidates.
* cumulative: A voter can cast multiple votes for multiple candidates.

I think it is important to at least give the _name_ of the system in the description so people know what is being described or referred to (and what to look up if they want more detail).  For example, there are other systems aside from approval voting where voters can cast votes for any number of candidates.

Also, if you do want to add more description after giving the name, you will need to be more careful.  For example, for cumulative voting, the description seems to suggest that voters can always mark the same candidate more than once.  But in the equal and even method (which I believe is the most common cumulative method used in US government elections), voters can choose each candidate (up to _n_ of them) at most once, not multiple times.  From a voter perspective, it is like n-of-m, just tabulated differently.

> What if a voter can select n candidates, but p will actually end up winning, where n != p?

Well, I don't know of any cases like that in the US (where n isn't 1), but I would be okay with removing that clause.

> Also, right now we only allow on VoteVariationType value per contest.

Okay, then I would tweak some of my suggestions.  For example, majority and plurality should probably be special cases of 1-of-m.

pkoms:
Two obscure points on `DistrictType`:

*I like @johnpwack's formulation of municipality, for a separate, pedantic reason: municipality is said to be any "...civil division which is not a town, city, village, or county." We'll eventually have to account for some unincorporated localities (e.g., Maine's unincorporated No 21 Twp, which is a distinct voting district), and by definition, an unincorporated locality is not a municipality, and I don't believe it's really a civil division in a formal sense either. Since I don't have a good word at hand for "incorporated or unincorporated thing that's smaller than a county and not otherwise covered", changing from civil to administrative division and generalizing the term a bit seems right.
*Defining "township" as "a town" is going to mystify some states, I believe mostly Midwestern, where those are two different types of legal entities. Can we make it consistent -- either "Township -- a township" or "Town -- a town"?

docs/xml/enumerations/vote_variation.rst
```diff
@@ -1,35 +1,32 @@
...
 | Name                 | Description                                                                      |
 |                      |                                                                                  |
 +======================+==================================================================================+
-| 1-of-m               |                                                                                  |
-+----------------------+----------------------------------------------------------------------------------+
-| approval             |                                                                                  |
+| 1-of-m               | When a voter can select only one candidate in a contest.                         |
 +----------------------+----------------------------------------------------------------------------------+
-| cumulative           |                                                                                  |
+| approval             | When a voter can select as many candidates as desired.                           |
```
cjerdonek: If you want to include the description, then I would say:

    Approval voting, a system that allows voters to select as many candidates as desired.

jdmgoogle: Personally I think saying that the "approval" value of `VoteVariationType` means "Approval voting" is redundant, unless the "Approval voting" text is a hyperlink to wikipedia. I'll welcome other comments on this one, though.

cjerdonek: It adds something because not every enum value in our list has the pattern: "value" translates to "[value] voting."  This is only true for some of them, so it adds something to point out which ones.

Also, it might seem obvious to us, but not all readers necessarily know that "approval voting" is a "thing" (with a defined meaning, etc).  They might just think it is a shorthand word we picked.

I'm also okay with hyperlinking the text "Approval voting" to the Wikipedia entry.

jdmgoogle:
@cjerdonek I added a bunch of links to Wikipedia to help explain the terms themselves without having to repeat the names in the descriptions.

cjerdonek:
@jdmgoogle Okay, it is good that the links are there, but one still has to click on the link to see that the system referenced by the enum goes by the name "approval voting," etc.  Is there some type of character limit in the formatting that is preventing you from wanting to include the text "Approval voting," etc?

Also, I had a number of other comments above that aren't addressed in the latest revision.  If it would help move the conversation along, I can propose language that addresses the concerns I raised.

cjerdonek:
For comparison, other sections repeat the name.  For example--

    | city                 | A city.
    | county               | A county.
    | ward                 | A ward.

There is nothing wrong with explicitly stating the concept being referenced for clarity.  This is the same approach I suggested in my [original comment](https://github.com/votinginfoproject/vip-specification/pull/171#issuecomment-105801727).

jdmgoogle:
True, but "city" and "county" are generally well-understood. I'd actually like to improve "ward" since that can have different meanings in different places.

If people really feel it's necessary to repeat the value in the description, I will make that change. Otherwise it simply seems like padding out the text in order to hit some imaginary target word count.

jdmgoogle:
As for your previous points the only one I see that's left unaddressed is the equal-and-even versus points method. The rest appear to be ones which @johnpwack need to address given that they come from 1622.2 and would need to be changed there as well.

cjerdonek:
I will suggest language then.

cjerdonek:
I created PR #184 to take a second pass at documenting the VoteVariation enum.  It builds on the discussion here and addresses the issues I raised, as well as some others I didn't specifically mention.

jungshadow:
Personally, I move to accept this PR as-is and move the rest of the `VoteVariation` discussion to @cjerdonek's PR. The `DistrictType` argument we can continue to have over the next few days and hopefully @johnpwack, @Josh-LACRRCC, @kennethmbennett, and others will chime in. Additionally, when we merge this, a new set of docs will be built on ReadTheDocs, which will allow us to get opinions from folks that don't monitor GitHub. Fair?

jdmgoogle:
Fair enough. Can we get another LGTM to submit this and move discussion of `VoteVariationType`-only over to #184?

jungshadow:
LGTM /cc @pstenbjorn @jktomer @cjerdonek

cjerdonek:
> Can we get another LGTM to submit this and move discussion of VoteVariationType-only over to #184?

Yes, LGTM if the changes to `vote_variation.rst` are left out.

jdmgoogle:
Merging and taking up the `VoteVariation` discussion in #184.

PR #176, opened 2015-05-26
Document ContactInformation

### LINKED ISSUE #175: Document ContactInformation
Finish documenting `ContactInformation`. I marked this also as a question, but it's more of a statement about `identifier`:

```xml
<xs:complexType name="ContactInformation">
  ...
  <xs:attribute name="identifier" type="xs:string" use="required" />
</xs:complexType>
```

Reading back through #144, I think we may have put the cart before the horse. I'd prefer @nomadaisy and/or @pkoms weigh in on this convention, since they're in charge of "CSV-ing" the spec. If they think this idea works the best for them...great and I'll shut up :) /cc @MariaDemocracyWorks

jungshadow:
Fixes #175 after merge. Will need one more update depending on the result of the discussion on #175.


jdmgoogle:
Looks good. For the `Hours` element, please add a note saying that this field is deprecated and will be going away in future versions.

jungshadow:
@jdmgoogle Fixed. Looking for one more LGTM from @pstenbjorn, @cjerdonek, @jktomer, @pkoms, @nomadaisy, or anyone else :)

pstenbjorn:
@jungshadow one comment, the `AddressLine` description is defined as the location information about physical addresses, will this not be used for mailing addresses too?  Aside from that LGTM

jungshadow:
@pstenbjorn Fixed. Just waiting for @pkoms or @nomadaisy to weigh in on #175 and I'll update/merge.

nomadaisy:
LGTM

jungshadow:
Added description of `identifier` in the preamble (NB: I didn't want to put it in the table as it feels slightly wrong, since the table data under the header "Repeats?" is "Attribute" and all other rows are elements).

jungshadow:
@cjerdonek I characterized `identifier` as being similar to a generic XML id, but, while the attribute is required, there's nothing enforcing uniqueness across the document or among `ContactInformation` elements. Does how I characterized it align with your thinking on this?

cjerdonek:
@jungshadow I'm glad I saw this. I haven't been following all of the PR's as closely as I was before.

So it looks like there is a typo / mistake in the spec.  The "identifier" attribute for ContactInformation should actually be optional rather than required.  It should function exactly as the identifier attribute for InternationalizedText (which is also optional).  Also, it doesn't need to be unique.

I meant to add something to the style guide about these "identifier" attributes since they all function the same way, but I never got around to it.  For now, I'll explain it here.

The "identifier" attribute is meant for use when using flat files.  For example, for a flat file providing ContactInformation objects, each row of the flat file can have a unique "identifier" field (unique across all rows in the flat file).  Then, when that ContactInformation object / row is incorporated into an XML feed, the "identifier" attribute should be included as well so people know from which row the flat file came.

The reason the identifier doesn't need to be unique across an XML feed is that it's possible for the same ContactInformation object / row to appear in multiple spots in the feed (since they are being in-lined in the feed as opposed to being listed in a global top-level node).

It might be clearer if "identifier" were renamed to something that made reference to flat files, e.g. "flat_file_id" or "flat_file_identifier".  Then it would be more obvious to people that the identifier attribute relates to flat files rather than something else.

jdmgoogle:
Maybe instead of "identifier" we used "tag" or "label"?

jungshadow:
I'd like to get this merged and this is starting to seem like an spec bug/change, so I'll update the copy on this to match @cjerdonek's thoughts on how this works and we can open a separate issue to debate on how this should work with the CSVs (attn: @pkoms and @nomadaisy). Cool?

jungshadow:
Assuming this works for all, can I get a LGTM? /cc @jdmgoogle, @jktomer, @pstenbjorn, @cjerdonek, @pkoms, @nomadaisy

pstenbjorn:
@jungshadow LGTM

jdmgoogle:
LGTM

PR #178, opened 2015-05-27
Added LatLng to the PollingLocation object.

### LINKED ISSUE #165: Include lat/long for polling locations
The second comment from the states:

> I would suggest the ability to submit “XY” latitude/longitude coordinates for polling locations for those who collect that information (NB: we use decimal degrees to seven significant figures). We find that using polling location coordinates is more reliable and more accurate than traditional address interpolation.

This did exist in the never-launched 4.0 specification. It may be worth adding this to `ContactInformation` as well rather than just polling location in case it could be useful for other elements (e.g. `ElectionAdministration`). To that point, does it make sense to extract the address "guts" of `ContactInformation` and reuse in `PollingLocation` (NB: I have a feeling we've probably had a lengthy conversation about this...but I don't remember)?

jdmgoogle:
This should fix #165.

### DIFF
--- a/docs/xml/elements/polling_location.rst
+++ b/docs/xml/elements/polling_location.rst
@@ -89,6 +89,8 @@ latitude and longitude values are measured in decimal degrees.
 |                       |                       |              |            |geocoding service.  |                       |
 +-----------------------+-----------------------+--------------+------------+--------------------+-----------------------+
 
+.. _`WGS 84`: http://en.wikipedia.org/wiki/World_Geodetic_System#A_new_World_Geodetic_System:_WGS_84
+
 .. code-block:: xml
    :linenos:
 
@@ -103,3 +105,4 @@ latitude and longitude values are measured in decimal degrees.
         <Source>Google Maps</Source>
       </LatLng>
    </PollingLocation>
+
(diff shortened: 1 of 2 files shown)


jungshadow:
Perfect.

jungshadow:
Pinging @pstenbjorn, @cjerdonek, @jktomer, et al for sign-off.

pstenbjorn:
Looks good to me.  Thanks

docs/xml/elements/polling_location.rst
```diff
@@ -50,12 +50,47 @@ The Polling Location object represents a site where voters cast or drop off ball
 |                       |                       |              |            |polling location is |the implementation is  |
 |                       |                       |              |            |an early vote site. |required to ignore it. |
 +-----------------------+-----------------------+--------------+------------+--------------------+-----------------------+
+| LatLng                | :ref:`LatLng          | Optional     | Single     |**LatLng** specifies|If **LatLng** is       |
+|                       | <pl-latlng>`          |              |            |the latitude and    |invalid or not present,|
```
jktomer: > **LatLng** specifies the latitude and of this polling location.

you accidentally a longitude

jktomer:
LGTM

jdmgoogle:
Thanks. Merging.

PR #180, opened 2015-05-27
Fix spelling of 'Latitude' in the spec.

### LINKED ISSUE #165: Include lat/long for polling locations

jdmgoogle:
This'll fix #165 for real.

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -493,7 +493,7 @@
       <xs:element name="LatLng" minOccurs="0" maxOccurs="1">
         <xs:complexType>
           <xs:sequence>
-            <xs:element name="Latitutde" type="xs:float" minOccurs="1" maxOccurs="1" />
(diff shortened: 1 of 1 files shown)


jktomer:
LGTM

jungshadow:
LGTM

Issue #183, opened 2015-05-28
Document VoteVariation enum


cjerdonek:
Document the VoteVariation enum.

cjerdonek:
Closed by PR #184.

PR #184, opened 2015-05-28
Address issue #183: document VoteVariation enum

cjerdonek:
This PR is a second pass at documenting the VoteVariation enum, building on the discussion in PR #171.  It addresses issue #183.


docs/xml/enumerations/vote_variation.rst
```diff
@@ -1,35 +1,43 @@
...
 +----------------------+----------------------------------------------------------------------------------+
-| cumulative           |                                                                                  |
+| approval             | `Approval voting`_, where each voter can select as many candidates as desired.   |
 +----------------------+----------------------------------------------------------------------------------+
-| majority             |                                                                                  |
+| cumulative           | `Cumulative voting`_, where each voter can distribute their vote to up to        |
+|                      | *N* candidates.                                                                  |
 +----------------------+----------------------------------------------------------------------------------+
-| n-of-m               |                                                                                  |
+| majority             | A 1-of-m method where the winner needs more than 50% of the vote to be elected.  |
```
jdmgoogle: There are no non-1-of-m elections which require a majority in the entire US?

docs/xml/enumerations/vote_variation.rst
```diff
@@ -1,35 +1,43 @@
...
 | Name                 | Description                                                                      |
 |                      |                                                                                  |
 +======================+==================================================================================+
-| 1-of-m               |                                                                                  |
-+----------------------+----------------------------------------------------------------------------------+
-| approval             |                                                                                  |
+| 1-of-m               | A method where each voter can select up to one candidate.                        |
 +----------------------+----------------------------------------------------------------------------------+
-| cumulative           |                                                                                  |
+| approval             | `Approval voting`_, where each voter can select as many candidates as desired.   |
```
jdmgoogle: Still duplicate wording IMHO. Paging @jktomer @jungshadow @pstenbjorn etc

docs/xml/enumerations/vote_variation.rst
```diff
@@ -1,35 +1,43 @@
...
+|                      | *N* candidates.                                                                  |
 +----------------------+----------------------------------------------------------------------------------+
-| n-of-m               |                                                                                  |
+| majority             | A 1-of-m method where the winner needs more than 50% of the vote to be elected.  |
 +----------------------+----------------------------------------------------------------------------------+
-| plurality            |                                                                                  |
+| n-of-m               | A method where each voter can select up to *N* candidates.                       |
 +----------------------+----------------------------------------------------------------------------------+
-| proportional         |                                                                                  |
+| plurality            | A 1-of-m method where the candidate with the most votes is elected, regardless   |
```
jdmgoogle: There are no non-1-of-m elections (e.g., approval, cumulative) which can be won via a plurality?

docs/xml/enumerations/vote_variation.rst
```diff
@@ -1,35 +1,43 @@
...
+| rcv                  | `Ranked choice voting`_ (RCV), where each voter can rank the candidates, and     |
+|                      | the ballots are counted in rounds.  Also known as instant-runoff voting (IRV)    |
+|                      | and the single transferable vote (STV).                                          |
 +----------------------+----------------------------------------------------------------------------------+
-| rcv-borda            |                                                                                  |
+| rcv-borda            | `Borda count`_, where each voter can rank the candidates, and the rankings are   |
+|                      | assigned point values.                                                           |
 +----------------------+----------------------------------------------------------------------------------+
-| super-majority       |                                                                                  |
+| super-majority       | A 1-of-m method where the winner needs more than some predetermined fraction     |
```
jdmgoogle: There are no non-1-of-m elections (e.g., IRV) in the US which require a super-majority?

docs/xml/enumerations/vote_variation.rst
pstenbjorn: @jdmgoogle is your concern that the documentation is too verbose if "approval voting" is the definition?  I agree that the enum value should read `approval`.

jktomer: I can't imagine caring about the distinction in question here enough to think about it for more than three seconds.

docs/xml/enumerations/vote_variation.rst
cjerdonek: Yes, there is single-winner RCV for example (in the final round of the tabulation).  But for that case, I would hope we want them to put "rcv" rather than "majority".

As for others, I don't know.  But the problem is that if there is, how would the jurisdiction know if they should label it "majority" or the other thing?  (When I asked, you said that feed creators were only allowed to choose one enum value per contest as opposed to multiple.)

Given that, I would rather us resolve the confusion for them (so it's obvious what enum value they need to choose) as opposed to leaving things ambiguous and so uncertain as to what what enum they're supposed to pick.

docs/xml/enumerations/vote_variation.rst
cjerdonek: Yes, you could say that the winners in approval voting and cumulative voting are also chosen by "plurality."  But my reply here is the same as [my reply](https://github.com/votinginfoproject/vip-specification/pull/184/files#r31302240) in the "majority" case:

Rather than leave things ambiguous (i.e. potentially multiple enum values being applicable to a given contest), I would rather the documentation give clear guidance as to which enum value should be chosen.  For example, for something like approval voting or cumulative voting, I'm betting we can all agree that we would want them to choose "approval" and "cumulative" respectively, rather than "plurality".

Also, just to be clear, by documenting what the enum name "plurality" stands for, that doesn't mean we need to be defining the English _word_ "plurality" in the abstract, general sense.  The enum name is just an abbreviation for the purposes of the VIP spec, and we are free to say what it stands for in the context of the spec (i.e. it doesn't need to mean plurality _literally_).

docs/xml/enumerations/vote_variation.rst
cjerdonek: Same as my previous two replies [here](https://github.com/votinginfoproject/vip-specification/pull/184/files#r31302240) and [here](https://github.com/votinginfoproject/vip-specification/pull/184/files#r31303001) in the "plurality" and "majority" cases.  I think we should resolve the ambiguity now rather than leave everyone else scratching their heads as to which enum value to choose when more than one is applicable.

cjerdonek:
I added a note to address the point that @jdmgoogle raised about how there could be non-1-of-m voting methods that are considered majority (in the general sense of the word), and similarly for plurality and super-majority.

Essentially, the note says that the descriptions should be taken as descriptions of what the enum values stand for in the context of the VIP spec, rather than as general definitions of the corresponding election terms (in cases where the enumeration name corresponds to an election term).

cjerdonek:
Regarding the "duplication" that @jdmgoogle mentioned, I think it's important that we include in the description the full name of the voting system being referred to (for example "approval voting" in the case of "approval", "borda count" in the case of "borda", and "ranked choice voting" in the case of "rcv").

jdmgoogle:
Taking up the `VoteVariation` discussion from #171.

I'd really like for @pstenbjorn to weigh in here, since he probably has a good sense of how election officials across the US will view/need these to be defined and clarified. I'll defer to his comments and experience.

I'd also like for @jktomer to weigh in from the data processing pipeline side of things, but that's a secondary issue, IMHO.

pstenbjorn:
@jdmgoogle the vast majority of electoral contests in the US are 1-of-m or n-of-m (although they wouldn't use that terminology) with a smattering of rcv and irv.  Most election officials are not particularly familiar with the lexicon of alternate vote methods so being descriptive in our definitions would be helpful.   So, in short, I agree with @cjerdonek that providing definitions which describe the electoral process associated with the enum value would probably be useful.

jdmgoogle:
@pstenbjorn Thank you. Do you have any specific comments on the wording @cjerdonek has used in this PR to describe the variation types? Or should we merge as-is?

pstenbjorn:
@jdmgoogle in general, I think these are fine descriptions.  One quibble is referring to ballot selections as `candidates`.  For example 1-of-m voting might be `A method where each voter may make up to one ballot selection from one or more options.`

jdmgoogle:
Okay. If @cjerdonek updates the wording to reflect ballot selections in general instead of candidates specifically, that's enough to get a LGTM.

Thank you @pstenbjorn and @cjerdonek 

(I heard from @jktomer this morning that other obligations may keep him from participating in the rest of the VIP discussions, hence the short-circuiting of the second half of my earlier request.)

cjerdonek:
Thanks a lot, guys.  Updated per @pstenbjorn's comment and merged.

Issue #201, opened 2015-06-01
Document Party element


jdmgoogle:
Document the `Party` element.

PR #214, opened 2015-06-02
Documentation for the Party element.

jdmgoogle:
This addresses #201.

### DIFF
--- a/docs/xml/elements/party.rst
+++ b/docs/xml/elements/party.rst
@@ -11,7 +11,7 @@ This element describes a political party and the metadata associated with them.
 |                     |                           |           |          |the party name.       |present, the implementation is |
 |                     |                           |           |          |                      |required to ignore it.         |
 +---------------------+---------------------------+-----------+----------+----------------------+-------------------------------+
-| Color               | HtmlColorString           | Optional  | Single   |The preferred display |If the field is invalid or not |
+| Color               | `HtmlColorString`_        | Optional  | Single   |The preferred display |If the field is invalid or not |
 |                     |                           |           |          |color for the party,  |present, the implementation is |
 |                     |                           |           |          |for use in maps and   |required to ignore it.         |
 |                     |                           |           |          |other displays.       |                               |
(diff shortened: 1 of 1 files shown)


docs/xml/elements/party.rst
```diff
@@ -1,35 +1,35 @@
...
++---------------------+---------------------------+-----------+----------+----------------------+-------------------------------+
+| Tag                 | Data Type                 | Required? | Repeats? |Description           |Error Handling                 |
+|                     |                           |           |          |                      |                               |
++=====================+===========================+===========+==========+======================+===============================+
+| Abbreviation        | xs:string                 | Optional  | Single   |An abbreviation for   |If the field is invalid or not |
+|                     |                           |           |          |the party name.       |present, the implementation is |
+|                     |                           |           |          |                      |required to ignore it.         |
++---------------------+---------------------------+-----------+----------+----------------------+-------------------------------+
+| Color               | HtmlColorString           | Optional  | Single   |The preferred display |If the field is invalid or not |
```
jungshadow: Do we define `HtmlColorString` elsewhere? I forget.

jungshadow:
Other than the one question, LGTM. /cc @pstenbjorn @cjerdonek

docs/xml/elements/party.rst
jdmgoogle: No. Not sure we do `TimeWithZone`, either. We should probably fix that. I'll file issues.

jdmgoogle:
Filed #215 to track documentation for `HtmlColorString`.

jungshadow:
@jdmgoogle Do you feel we should just treat `HtmlColorString` like `LanguageString` in that we folded the latter into `InternationalizedText`? I'm not against tracking it elsewhere, but it may be overkill to create a new section of types if their uses are isolated.

jdmgoogle:
I'd be okay with that; my only concern is that if we end up using it elsewhere in the future we'll need to move it, but I guess that's a relatively minor issue.

I'll make the change by adding `HtmlColorString` to `Party` and `TimeWithZone` to `HoursOpen`.

jdmgoogle:
Latest commit addresses `HtmlColorString` for #215.

jungshadow:
LGTM yet again.

jdmgoogle:
Thanks.

@pstenbjorn @cjerdonek  @nomadaisy @pkoms

pstenbjorn:
LGTM

PR #221, opened 2015-06-04
Added Name element to Election

### LINKED ISSUE #219: Add a Name element to Election
Late addition to the spec, but one of the things we continue to run into is what a particular election is called. All the ancillary information in `Election` is important, but doesn't communicate a user-friendly name for the upcoming event. My thoughts:

* It should be Optional for now, but move to **Required** after we upgrade folks from 3.0 to 5.0;
* The type should be `InternationalizedText`;

The wrinkle is that we roll lots of elections into a single feed, so, for lower-level elections in states that don't consolidate, the name of the election may be hard to figure out or obnoxiously long if the result is a concatenation of all the events.

I'd like to be able to shoehorn this thing in before we push this out the door. Thoughts?

jungshadow:
Closes #219


jdmgoogle:
LGTM

pstenbjorn:
LGTM

Issue #179, opened 2015-05-27
Keep sample feed synced with spec


jdmgoogle:
This is an issue that should be kept open until 5.0 is officially released. It is here to remind us to keep the sample feed synced with the spec itself. Any PRs which fix the sample feed from earlier breakages should reference this issue.

jdmgoogle:
Items that need to be added to the sample feed:

1. Ballot measure contests;
2. Retention election (Utah 2014 sample data);
3. Judicial elections (compare/contrast LA elections with item (2));
4. Internationalized text;
5. Early voting sites;
6. Combination early voting site/election-day voting site;
7. Drop boxes;
8. Mail-only precincts;
9. A `LatLng` in a  `PollingLocation`.

jungshadow:
@jdmgoogle Hope you don't mind that I changed your numbered list into a todo list. Just thought it might be easier to tick off the things when they were finished :)

jdmgoogle:
Moving this to "5.0 Documentation" since it has to do with the sample feed. The only item left is for me to examine how LA's judicial elections may be different than retention elections.

jdmgoogle:
Barring any last-minute changes because of #268 (which is highly unlikely) this will get closed out soon.

PR #223, opened 2015-06-04
Add examples to sample feed

jdmgoogle:
The examples added include:
1. `LatLng` for several polling locations;
2. A `BallotMeasureContest`;
3. An early vote site which is also an election-day site; and
4. An `HoursOpen` example which spans time zone changes (EST vs EDT).

This addresses some of the requests in #179. There will be at least one more PR after this one which adds the other requests from #179.


pstenbjorn:
@jdmgoogle these look good to me.

jungshadow:
LGTM

Issue #192, opened 2015-05-29
Make sure "identifier" attribute is present and correct as needed


cjerdonek:
We need to make sure that a flat-file "identifier" attribute is present (as an optional attribute) for all types whose objects are instantiated in-line (i.e. not defined in a top-level node of the XML document).  This includes things like ContactInformation, Term, etc.  The reason for this, again, is to allow tracking nested objects back to their corresponding row in a flat file when flat files are used.

cjerdonek:
See the related issue #149, which is to document this pattern in the style guide.

cjerdonek:
Closed by PR #229.

PR #229, opened 2015-06-05
Address issue #149 and issue #192: flesh out non-primary object ID.

cjerdonek:
This is a first crack at addressing issue #149 and #192.

I'm addressing them together because the documentation explains the XSD changes, so it seems more helpful to commit them together.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -70,7 +70,7 @@ Markdown, etc).
   instances of the type occur only as immediate children of the "VipObject"
   root object.  We use this as the requirement because this is a simple
   way to characterize whether instances of the type are defined "in one
-  place" in an XML feed (and so cannot occur twice).
+  place" in an XML feed (and so not at risk of occurring twice).
 
 * A primary key should be an attribute with type "xs:ID" and name "id".
   For example,
@@ -101,12 +101,21 @@ Markdown, etc).
 
 A complex type should have an optional attribute named "csvRowId" of type
 "xs:string" if (1) instances of the type are represented by rows of a flat
-file when using flat files and (2) the type does not already have a primary
+file when using flat files, and (2) the type does not already have a primary
 key attribute.  The attribute looks like this when present:
 
-    ```xml
-    <xs:attribute name="csvRowId" type="xs:string" />
-    ```
+```xml
+<xs:attribute name="csvRowId" type="xs:string" />
+```
+
+The "csvRowId" attribute is useful for troubleshooting, etc. because it
+lets one trace data in an XML feed that might not otherwise be traceable
+to the row of the flat file from which it came.
+
+The attribute is not a primary key relative to an XML feed because it is
+possible for the same attribute value to occur in more than one place in
+the feed (e.g. if the same "Term" instance occurs as an element of more
+than one "Office" instance).
 
 
 ### Element/Attribute Property Ordering
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -64,7 +64,7 @@ Markdown, etc).
   ```
 
 
-### Primary Key Attribute: `id`
+### Primary Key ID Attribute: `id`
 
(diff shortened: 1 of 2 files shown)


cjerdonek:
I switched the attribute name back to something more generic since it can be useful even when flat files are not used.  I'm not wedded to "objectId" or "identifier" though.

STYLEGUIDE.md
```diff
@@ -91,6 +97,28 @@ Markdown, etc).
     ```
 
 
+### Object ID Attribute: `objectId`
```
jdmgoogle: Proposal: instead of `objectId` (which could cause confusion with 1622.2), how about `label`? To me, at least, it's intuitive that multiple objects can have the same label.

jdmgoogle:
Thanks for writing this up. To echo my inlined comments, I propose we use the attribute name `label` instead of `objectId`.

Thoughts?

jungshadow:
I'll second `label`, since I agree that `objectId` might cause confusion. /cc @pstenbjorn, @pkoms, @nomadaisy since this falls into your realm.

cjerdonek:
Hmm, do you have any suggestions other than label?  The thing is that it really is an ID for the object.  It's just that multiple _copies_ of the same object can appear in the feed, so it can't technically be an attribute of type "xs:ID" in the sense of XML.

How about "internalId" or "internalIdentifier" (which has a nice parallel with externalIdentifier)?

nomadaisy:
I also like label or internalId

cjerdonek:
I changed the name away from "objectId" to address the issue raised by @jdmgoogle and @jungshadow of possible confusion with 1622.2.  In the update I went with "internalId" (which @nomadaisy is also okay with) because I feel it better communicates the idea that this is still in fact an ID for the underlying object.

jdmgoogle:
Circling back to this, I'm not a fan of `internalId`. So far we have 3 people who support `label`, 2 who support `internalId`, and 1 ( ::cough:: @jungshadow ::cough:: ) who has yet to weigh in on `internalId`.

Also paging @pstenbjorn and @jktomer.

pstenbjorn:
I am okay with either label or intenalid.

jungshadow:
Yeah...`label`'s fine. `internalId`'s also fine. Prefer `label`.

jdmgoogle:
@cjerdonek It appears there's general consensus/support for `label`.

jdmgoogle:
Ping @cjerdonek.

cjerdonek:
Okay, I can change to label.  Does this still need two "approvals" once that change is made?  I'm not sure anyone has weighed in on the substance of the patch.

jdmgoogle:
Once the s/internalId/label/g change has been made, LGTM.

cjerdonek:
PR updated.

cjerdonek:
I can merge this once someone gives the go-ahead (it's unclear to me if there has been a second LGTM).

jdmgoogle:
Thanks, @cjerdonek.

One last thing: what are people's opinions on making these labels required? Or should we keep them optional?

cjerdonek:
The main reason I didn't mark them required in the spec is that (if I understand correctly) not everyone provides data as flat files.  If flat files aren't used, then it wouldn't be obvious what value to use (e.g. maybe not every organization would have internal "ID's" for the objects).

So to summarize, I think _informally_ they should be required (when flat files are used).  But this can't be part of the spec because the spec isn't privy to how the data is being provided.

jdmgoogle:
That lines up with my thinking, too.

LGTM.

cjerdonek:
On the other hand, if XML is being used, I could see it helping to be able to reference the objects unambiguously, for example by saying "hey, your objects with label foo are messed up," as opposed to having to describe those objects indirectly (e.g. by line number, by contents that might not be unique, by the ID containing element, etc).

jdmgoogle:
Paging @jungshadow

jungshadow:
LGTM

Issue #230, opened 2015-06-05
Synch VoteVariation with latest 1622.2


cjerdonek:
VSSC 1622.2 came out with their updated draft today (version 43, see [here](http://grouper.ieee.org/groups/1622/groups/2/)), so this issue is to synch our spec with their latest draft (in the places that we are mirroring 1622.2).  I mentioned that we might need to do this during our hangout meeting yesterday (on Wed, June 3).

PR #231, opened 2015-06-05
Address issue #230: synch with latest 1622.2

cjerdonek:
This PR is to address issue #230: synch with latest 1622.2.

The initial commit handles VoteVariation.  There may be other things that need updating on the VIP side, but I haven't yet checked exhaustively.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -73,7 +73,7 @@ Markdown, etc).
     <xs:attribute name="id" type="xs:ID" use="required" />
     ```
   Note that this departs from the style used by [VSSC 1622.2][vssc_1622],
-  where the name is "object_id".
+  where the name is "ObjectId".
 
 * A reference to a primary key should have type "xs:IDREF".  The name
   of the element or attribute should normally end with the name of the type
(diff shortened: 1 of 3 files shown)


jdmgoogle:
Thanks for keeping an eye on this.

There are a few other small changes, including

1. the renaming of `OrderedContest.BallotSelectionId` to `OrderedContest.OrderedBallotSelectionId`;
2. Adding `FullName` to `Person`;
3. Adding "town" to `DistrictType`; and
4. Adding "measure" to `VoteVariationType`.

There's also an `OfficeGroup` element, which appears to create a hierarchy of `Office` elements in the same way that 1622.2's `GPUnit` allows for hierarchy. I'm not sure we want to/need to copy that over, though.

I'll take care of items 1-3 in another pull request, and check with @johnpwack about the motivation behind item 4. I'll also file a discussion issue for the `OfficeGroup` element, but I'm hesitant to add something this late in the game.

jdmgoogle:
As for this PR: LGTM.

jungshadow:
LGTM

@jdmgoogle I'll wait until you create another issue to talk about the 1622 updates, but, in short, I think I agree with your thoughts on the `OfficeGroup` structure.

cjerdonek:
Thanks.

Issue #249, opened 2015-06-16
Make tables in docs viewable without horizontal scrolling on Desktop


cjerdonek:
It would be helpful if we can somehow make the tables in the documentation fully visible without having to scroll the table horizontally.  I'm attaching a screen shot of how a table currently looks:

![vip](https://cloud.githubusercontent.com/assets/355822/8174943/136a2d9a-1398-11e5-9e81-536f20efddb4.png)

jungshadow:
@cjerdonek This is something I'm planning to tackle. I'm a little concerned that we may be pushing the boundaries of the template by using (somewhat intractable) 6-column tables. The tables are currently a problem for the generated PDF. We may eventually need to condense the tables into 5-columns (not my favorite option) or approach writing the docs in a more long-form way (e.g. [the python documentation](https://docs.python.org/3.4/library/difflib.html#difflib.SequenceMatcher.find_longest_match)).

cjerdonek:
Personally, I think HTML tables are fine (e.g. the ones you see in the v3 docs here: http://votinginfoproject.github.io/vip-specification/ ).

In the short term, I was hoping that this issue could be addressed by a simple change (e.g. a CSS change to allow text wrapping within table cells).  In the longer term, I was going to recommend generating the docs programmatically (something I've mentioned before).  It doesn't seem like it should be necessary to edit HTML or reST tables by hand.  If the tables are generated programmatically, then things like table reformatting could be done in one place (e.g. in the script or template driving the auto-generation).

jungshadow:
@cjerdonek Sorry, Chris. I might not have been entirely clear. Personally, I like the look of the tables, but we're breaking through the hard "physical" boundaries of the theme. When on smaller screens/devices, there just isn't enough space for the a 6-column table to exist without scrolling, especially with names like `BallotMeasureType` or `InternationalizedText` which can't wrap or be abbreviated.

The change is going to be in the CSS and will be theoretically simple, but it's still not going to solve the overall problem. In addition, it doesn't fix the PDF problem at all, but we'll cross that bridge when we come to it.

jungshadow:
Doing a bit of experimentation, this is the result of some of the lengthier tables (_**NB:** it's hard to see, but the table expands beyond the established margins of the page_):

![table_example](https://cloud.githubusercontent.com/assets/85972/8192573/12f230b0-143d-11e5-8d78-417e2a755d0a.png)

cjerdonek:
[...]

PR #250, opened 2015-06-18
Change table scroll behavior

jungshadow:
This should address #249 to a limited degree. While it does expose the entire table in the main view, the tables are effectively too wide to fit within the confines of the theme. More work will have to be done in a future PR.

### DIFF
--- a/docs/_static/theme_overrides.css
+++ b/docs/_static/theme_overrides.css
@@ -1,0 +1,11 @@
+/* override table width restrictions
+ * modified solution from https://github.com/snide/sphinx_rtd_theme/issues/117#issuecomment-41606147
+ */
+.wy-table-responsive table td, .wy-table-responsive table th {
+    white-space: normal !important;
+    padding: 4px 8px !important;
+}
+
+.wy-table-responsive {
+    overflow: visible !important;
+}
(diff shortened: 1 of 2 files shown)


jungshadow:
@cjerdonek @jdmgoogle @pstenbjorn Let me know what y'all think.

pstenbjorn:
@jungshadow from what I can tell, this LGTM

cjerdonek:
Thanks, @jungshadow.  My only comment is to add a code comment to the CSS with a link to issue #249 so people can know more easily why the extra CSS is there.  Other than that, LGTM.

jungshadow:
Thanks, @cjerdonek and @pstenbjorn!

@cjerdonek I don't think the comment is necessary, since I was planning to make this change in the beginning. Plus, it seems odd to have links to the repository history/issues inside the code. Seems a bit like "institutional underpants" to borrow a phrase from a friend of mine (i.e. the inner-workings of the decision are exposed to the public who don't need/want the details) :)

jungshadow:
Sorry, @jktomer, I forgot to flag you on this and that @jdmgoogle was going to be out. What do you think (see screenshots on #249)?

jktomer:
LGTM

Issue #233, opened 2015-06-05
Update CONTRIBUTING with instructions for running Sphinx


jungshadow:
As advertised above, this is to help contributors build the documentation prior to contributing.

PR #251, opened 2015-06-18
Added instructions for installing Sphinx

jungshadow:
This should address some of #233. I don't have a PC to supply non-*nix-based instructions, but anyone else should feel free to contribute updates to this PR.


jungshadow:
@jktomer @jdmgoogle @pstenbjorn @cjerdonek What do y'all think?

pstenbjorn:
LGTM.  As far as Sphinx for PC the instructions are pretty much the same, it expects python to be installed and then the command is the pip install as well.

jktomer:
LGTM

Issue #255, opened 2015-06-28
Auto-generate tables in documentation


cjerdonek:
This issue is to auto-generate the reST tables in the documentation from structured data, which is something I've suggested in the past.

I think this will open up a lot of possibilities for us.  For one, it will make the reST tables a lot easier to manage because we will be able to manage their appearance and formatting in one place.  (It is also _way_ easier to edit a YAML value than to edit a value in a reST cell!)

Some other advantages: it will reduce the amount of information we need to update by hand since everything will be more DRY.  For example, strings like "If the field is invalid or not present, the implementation is required to ignore it." can be set globally in the script that auto-generates the tables.  Auto-generating will also make it straightforward to "refactor" the tables (e.g. by adding or eliminating a new column).

A couple other things: it would open up the possibility of making alternative renderings of the tables (e.g. HTML for responsive design, etc, if the reST tables are too limiting).  Finally, it will be a lot easier to do global fixes and globally search the text values since we can just iterate over YAML data (i.e. without needing to parse anything, etc).

cjerdonek:
I created PR #256 for this.

cjerdonek:
To illustrate a bit more, you can see how values like "`Optional`", "`**Required**`", and "If the field is invalid or not present, then the implementation is required to ignore it." are set globally [here](https://github.com/cjerdonek/vip-specification/blob/f69a742468f381f7866c3539358a829146993490/scripts/vippy/rest.py#L75).  I noticed lots of variation in the past phrasing, a lot of which is made consistent now (though more can be done).

cjerdonek:
I just posted PR #273 for Part 2 to address this issue.

jdmgoogle:
Is there anything left on this particular issue which is *not* covered by #275?

cjerdonek:
The only things I can think of are:

1. there was a request to document the Python code a little better (e.g. more docstrings), and
2. the error-related values in the YAML can be made more DRY (e.g. similar to what was done in PR #274 around error messages with the phrase, "ignore the ``parent`` element containing it").

But I would recommend creating separate issues for those rather than keeping this issue open.

jdmgoogle:
I filed issues #279 and #280 to keep track of those. Closing this now.

PR #256, opened 2015-06-28
Address issue #255: Auto-generate tables in documentation

cjerdonek:
This addresses issue #255: "Auto-generate tables in documentation."


CONTRIBUTING.md
```diff
@@ -29,13 +29,32 @@ While the following information shouldn't be considered hard-and-fast rules, the
...
+## Dev Environment
+
+This section explains how to set up your local development environment for
+contributing.  First, [install Python][python_install].  We recommend
+installing the latest stable version of Python 3 (Python 3.4.3 as of June 2015).
+
+We also recommend setting up a virtual environment (e.g. using
+[virtualenv][virtualenv]) prior to installing dependencies.
+
+To install dependencies, open a terminal window and run:
```
jdmgoogle: Please include a link to the `pip` page.

scripts/vippy/common.py
```diff
@@ -0,0 +1,266 @@
...
+TAG_KEY_DESCRIPTION = 'description'
+TAG_KEY_ERROR_HANDLE = 'error'
+
+DEFAULT_TAG_VALUES = {
+    TAG_KEY_REQUIRED: False,
+    TAG_KEY_REPEATING: False,
+}
+
+ENUMERATIONS = set([
+    'BallotMeasureType',
```
jdmgoogle: Perhaps not in this PR, but in a future one we should update it to pull this set (and the SIMPLE_TYPES set) from the XSD itself.

jdmgoogle:
Generally this looks good.

@jungshadow @nomadaisy @pstenbjorn please take a few minutes to look this over. Thanks.

scripts/vippy/common.py
cjerdonek: Agreed.  For now, I can at least update the PR to get the list from the YAML files themselves (which it is already reading).  The two simple types I'll keep hard-coded though since those don't have YAML files.

cjerdonek:
Okay, @jdmgoogle, I addressed both of your comments, and merged the latest vip5 branch.

(Re: your second comment, I was able to refactor away the use of both module variables `ENUMERATIONS` and `SIMPLE_TYPES`.)

jdmgoogle:
@cjerdonek Okay, this looks really good.

Two quick things:

1) (Obviously) we need to wait for @jungshadow @pstenbjorn or @jen-tolentino to weigh in for the second LGTM.

2) Please hold off on merging this in to the main `vip5` branch until later today. There's something we need to sort out on our end first.

Thanks, and this looks great.

jungshadow:
I don't want to take away from all the good work you've done here, @cjerdonek, and I appreciate the effort you put into this.

Philosophically, I'm concerned about the inclusion of another moving part in the the documentation development process, which could make it even more daunting for prospective VIP participants. If we continued promoting this script as the means to generate the docs, wouldn't we be wedded to tables until someone wrote another output for it? How easy would it be to change in the future? Who would support it if no one in our current group had the bandwidth?

It would be great to give folks an option to use these scripts, but not make it mandatory. This code should live in its own repository and we can link to its potential use in the documentation. I hope that makes sense.

docs/data/enumerations/district_type.yaml
```diff
@@ -0,0 +1,46 @@
...
+- _name: judicial
+  description: A judicial district.
+- _name: municipality
+  description: A civil division which is not a town, city, village, or county.
+- _name: national
+  description: The United States.
+- _name: school
+  description: A school district.
+- _name: special
+  description: A `special-purpose district`_ that exist separate from general-purpose
```
jdmgoogle: How does the YAML handle/define links such as this one?

scripts/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2015, Christopher Jerdonek
+All rights reserved.
```
jdmgoogle: We (Pew and Google) are sorting out the licensing issue for all of VIP. It will be an OSI-compliant license, so once we finalize that we will put a copy of that license in the main repo. Once we've done that, please update this file to point to the main LICENSE file.

scripts/vip.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (c) 2015, Christopher Jerdonek
+# All rights reserved.
```
jdmgoogle: This should also be updated to contain a link to the appropriate top-level OSI-approved license once we've checked it in.

jdmgoogle:
So I have a few comments about the idea behind the PR as well as the PR itself.

1. After more discussion with @cjerdonek, I'm okay with this being a main part of the VIP repo and not just an optional external repo that users can also fork/pull in when they want to generate or update the documentation.
2. As @cjerdonek pointed out to me in email, in the worst case -- where we decide we no longer want to do this -- we can fall back on editing the RST tables by hand.
3. In the end it may be easier to break this into two PRs: one which contains the code and another which contains the updated documentation.
4. The code may be relatively straightforward, but a few more comments -- even at a high level -- explaining what each module does would be helpful, along with a one-liner for each function.
5. The YAML should also include the ability to define the top-level text for an element/enumeration, as well as information about links contained in the element descriptions.

docs/data/enumerations/district_type.yaml
cjerdonek: @jdmgoogle The processing of these strings is "dumb" in the sense that the strings are included in the resulting reST file as is.  So the links will work in the resulting reST provided that the hyperlink targets are also defined.

In this case, the reST looks like this in the generated table--

    | school         | A school district.                                 |
    +----------------+----------------------------------------------------+
    | special        | A `special-purpose district`_ that exist separate  |
    |                | from general-purpose districts.                    |
    +----------------+----------------------------------------------------+
    | state          | A state, district, commonwealth, or U.S.           |

And the table is included as follows in the reST:

    .. include:: ../../tables/enumerations/district_type.rst

    .. _`special-purpose district`: http://en.wikipedia.org/wiki/Special-purpose_district
    .. _town: http://en.wikipedia.org/wiki/Town#United_States
    .. _`Wikipedia article`: http://en.wikipedia.org/wiki/Town#United_States

As you can see, the needed hyperlink target appears after the inclusion of the table.

I also checked that Sphinx runs without errors when building the HTML, and also that the links work when browsing locally to that generated HTML.

scripts/LICENSE
cjerdonek: Thanks.  I'd be happy to do that.

cjerdonek:
@jdmgoogle I'd be happy to make those suggested changes (i.e. bullets 3, 4, and 5 in [this comment](https://github.com/votinginfoproject/vip-specification/pull/256#issuecomment-121746183)).  I won't be able to get to it this week though.  I can also "squash" the commits so that each PR will only have one commit, instead of having the dirtier history.

docs/data/enumerations/district_type.yaml
jdmgoogle: @cjerdonek I should have phrased it better. What I was trying to get at was the idea of adding some sort of "links" section at the bottom of the YAML, so that way the whole RST content (including the table and the links) would be generated from the YAML. E.g.,

```
links:
  _name: special-purpose district
  url: http://en.wikipedia.org/wiki/Special-purpose_district
  _name: town
  url: http://en.wikipedia.org/wiki/Town#United_States
  _name: Wikipedia article
  url: http://en.wikipedia.org/wiki/Town#United_States
```

cjerdonek: Sure, I could do that.

jungshadow:
This can't be merged until we have a final license for the spec, but this looks good. It needs a bit more documentation around what folks need to update–since it's not pure Sphinx and reST–and the commits need to be squashed, but I'm good with including this.

cjerdonek:
Great to hear, @jungshadow.  Thanks a lot.  Do you have an ETA on the license so I can have an idea of when I should aim to update the PR by (and improve the documentation, etc)?  (The ETA needn't be exact, just approximate.)

jungshadow:
@cjerdonek We're hoping to hear something final by the end of this week, but I'll let you know if that deadline changes.

cjerdonek:
See PR #272 which supersedes this one (though we should keep this one around at least temporarily for the comments).

cjerdonek:
Closing this as it was superseded by PR #272.

PR #259, opened 2015-07-01
Remove NonHouseAddress and fold its elements into the parent StreetSegment

jdmgoogle:
See $SUBJECT.

This involved much adjusting of the RST tables.

PR #257, opened 2015-06-29
Fix the sample feed (closing tag on IncludesAllAddresses is misspelled).

jdmgoogle:
See $SUBJECT.

PR #260, opened 2015-07-07
Fix the sample feed.

jdmgoogle:
I thought this had gotten fixed in #257, but apparently I undid it with #259. Oops.


jdmgoogle:
As this is literally a one-character change, I'm just going to merge this back in.

Issue #242, opened 2015-06-08
TTY phone numbers


cjerdonek:
I was helping someone implement ElectionAdministration for San Francisco, and I noticed that there doesn't seem to be a way to designate a TTY phone number.  See SF's "Contact Us" page [here](http://www.sfgov2.org/index.aspx?page=3335) for an example of a TTY phone number.

Currently, ContactInformation just has phone number elements for "Fax" and general "Phone," but no way to designate a phone number as TTY:

```xml
<xs:complexType name="ContactInformation">
  <xs:sequence>
    <xs:element name="AddressLine" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
    <xs:element name="Email" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
    <xs:element name="Fax" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
    <xs:element name="Hours" type="InternationalizedText" minOccurs="0" />
    <xs:element name="HoursOpenId" type="xs:IDREF" minOccurs="0" />
    <xs:element name="Name" type="xs:string" minOccurs="0" />
    <xs:element name="Phone" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
    <xs:element name="Uri" type="xs:anyURI" minOccurs="0" maxOccurs="unbounded" />
  </xs:sequence>
  <xs:attribute name="identifier" type="xs:string" use="required" />
</xs:complexType>
```

Also, if you list multiple "Phone" elements (e.g. a second number for TTY), there doesn't seem to be a way to indicate how the second phone number is different.

jdmgoogle:
Thoughts, @jktomer @jungshadow @pstenbjorn?

cjerdonek:
Here is another data point (from [CA SOS web site](http://www.sos.ca.gov/administration/contact-information#elections)):

    Phone Numbers
    Main Number: (916) 657-2166
    TTY/TDD: (800) 833-8683
    Voter Hotline: (800) 345-VOTE (8683)
    Elections Division Fax: (916) 653-3214

jktomer:
This is good to support. We should add a TDDNumber to ContactInformation (best not to wade into the quagmire of making things more generic).

jdmgoogle:
Closed by #261.

PR #261, opened 2015-07-07
Annotate strings with a purpose

jdmgoogle:
This is a rough attempt to address #242. The idea behind it is to have strings with optional attributes named `purpose` that allow the feed creator to indicate that an email address, phone number, or fax number is for a specified purpose. The obvious use here is for TTY/TDD phone numbers, but can also be used to indicate different divisions within an organization.

Feedback welcome.

### DIFF
--- a/sample_feed.xml
+++ b/sample_feed.xml
@@ -908,7 +908,7 @@
       <AddressLine>Washington, DC 20006</AddressLine>
       <Email>president@whitehouse.gov</Email>
       <Phone>202-456-1111</Phone>
-      <Phone purpose="TDD">202-456-6213</Phone>
+      <Phone annotation="TDD">202-456-6213</Phone>
       <Uri>http://www.whitehouse.gov</Uri>
     </ContactInformation>
     <FirstName>Barack</FirstName>
@@ -2187,7 +2187,7 @@
   <!-- Street segments -->
   <StreetSegment id="ss999999">
     <City>Charlottesville</City>
-    <IncludesAllAddresses>true</IncludeAllAddresses>
+    <IncludesAllAddresses>true</IncludesAllAddresses>
     <OddEvenBoth>both</OddEvenBoth>
     <PrecinctId>pre99999</PrecinctId>
     <State>VA</State>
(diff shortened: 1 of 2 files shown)


cjerdonek:
A few comments:

1. I would be concerned people would start abusing this by putting in descriptive phrases, sentences, or paragraphs even.  Is that allowed, and would we be encouraging that?
2. In the case of TTY phone numbers, people wouldn't be able to know programmatically something is a TTY number.  For that, we would need the annotation to be an enum (or else have a new tag as @jktomer suggested).
3. In terms of the name, "PurposeString" makes it sound like the tag _contents_ represent a purpose, when it is a string _with_ a purpose (i.e. "purposed string").  AnnotatedString?

jdmgoogle:
In order:

1. It'd be possible to define a `ShortString` object with a `maxLength` of something reasonable (e.g., 10-15 characters) and make `purpose` be of type `ShortString` instead of `xs:string`. This would cut down on people Dickens-ing their feeds.
2. It would be more heuristic instead of straightforward, yes. The idea behind this, though, is for feed creators to be able to annotate *any* phone number, email address, or other contact channel with a small piece of metadata.
3. True. AnnotatedString might work. This PR was more about generating discussion than a fully-formed solution.

jungshadow:
Responding to @jdmgoogle's responses:

1. I like the `ShortString` approach.
2. Agree with @jdmgoogle. We should restrict in the future, but currently "we don't know what we don't know" about what might be annotated. This has been the same approach we've taken with VIP, writ large (e.g. flexibility first, then restrict based on the data we have).
3. Agree with @cjerdonek. `AnnotatedString` feels more...right :)

jdmgoogle:
I'll update the PR to rename it to `AnnotatedString` and to cap the length at 16 characters.

jdmgoogle:
Updated. PTAL.

jungshadow:
A thought just occurred to me and I may be channeling @jktomer's concerns at this point. Is 16 characters enough to be universal (e.g. thinking of the hypothetical annotated email addresses you mention above)?

That said, I'm going to consciously forget I even mentioned that and focus on the immediate use case of phone numbers and say LGTM.

cjerdonek:
LGTM, too

jdmgoogle:
As a side note, this will fix #242.

PR #265, opened 2015-08-17
Update sample feed to use 'label' instead of 'identifier'.

jdmgoogle:
Also update comment in spec XSD to reflect identifier ~> label change. This is part of the #179 tracker.


cjerdonek:
LGTM

jungshadow:
The fix looks good, but looks like [the documentation](https://github.com/votinginfoproject/vip-specification/blob/issue/179-label-fix/docs/xml/elements/contact_information.rst#contactinformation) still needs an update in the preamble (i.e. change `identifier` to `label`). Should this also be included in the changelog?

jdmgoogle:
Good catch.

I plan to update HISTORY.md just prior to release. I filed #266 to track this.

jungshadow:
LGTM

jdmgoogle:
Thanks. Merging.

PR #272, opened 2015-09-01
Address issue #255: Take Two

cjerdonek:
This PR addresses issue #255 and is an improved version of PR #256.

This PR doesn't yet address all of the suggestions that were made in the comments to PR #256, but I wanted to get this PR posted as soon as possible to help meet the deadline, etc.

Note that this PR doesn't DRY up the YAML data quite as much as PR #256 since PR #256 involved some manual work mixed in with the automated script.  In this PR, however, everything is automated.

An example of something not handled by the current PR is the following:

> If the field is invalid or not present, the implementation is required to ignore the ``Office`` element containing it.

In the other PR, error-handling strings like this were generated automatically using a template (aka Python format string).  But in this PR, strings like this are "hard-coded" in the YAML.  It is slightly harder to convert these automatically since the word for the parent type varies.  This DRY-ing up can be done in a subsequent PR.

Also, this PR doesn't actually _run_ the parsing, YAML generation, and new reST table generation.  That can be done in a PR immediately after this, which was something that was requested.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,23 +51,91 @@ dev branch (called 'vip5' in the vip-specification's case, but it may change to 
 future), a stable branch (called 'master' in most cases), and, if necessary, a documentation branch
 (called 'gh-pages' if hosted on GitHub).
 
-# Installing Sphinx
-In order to build the documentation, you must install [Sphinx](http://sphinx-doc.org), which
-involves first [installing python](https://www.python.org/downloads/). Once python is installed,
-you should have access to the python package installer, `pip`. Open a terminal window and enter the
-following commands:
+
+## Documentation Format
+
+VIP writes its documentation in a format called [reStructuredText][reST]
+(aka reST).  reST is a markup language used mainly for documentation.
+VIP chose reST because of its enhanced formatting capabilities.
+For example, reST supports tables.
+
+The documentation that VIP exposes to the public is in HTML.  VIP generates
+its HTML documentation from the reST files automatically using a process
+called "building" the documentation.  Instructions for building are below.
+
+One problem with reST is that tables in reST are hard to edit and maintain
+by hand.  For this reason, VIP stores the tabular documentation data in
+separate files in an open data format called [YAML][YAML]
+(specifically [YAML version 1.1][YAML_1.1]). YAML files are human-readable and
+easy to edit by hand.  The YAML files are stored in the
+[`docs/yaml`](docs/yaml) directory.
+
+
+## Dev Environment
+
+This section explains how to set up your local development environment for
+contributing.  First, [install Python][python_download].  We recommend
+installing the latest stable version of Python 3 (Python 3.4.3 as of June 2015).
+
+We also recommend setting up a virtual environment for the repo (e.g. using
+[virtualenv][virtualenv]) prior to installing dependencies.
+
+Use [`pip`][pip] to install dependencies, which comes with Python 3.4+
+(and is installed automatically when creating a virtual environment).
+Open a terminal window and run:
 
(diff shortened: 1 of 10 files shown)


cjerdonek:
By the way, as a check, I ran a "diff" against the YAML files in the previous PR, and it mostly checks out except for some minor differences that result from changes since then, as well as things like error strings involving the "parent" type.

CONTRIBUTING.md
```diff
@@ -51,23 +51,103 @@ dev branch (called 'vip5' in the vip-specification's case, but it may change to
...
+[http://127.0.0.1:8000](http://127.0.0.1:8000) to see the documentation.
+
+
+## Updating Tables in the Documentation
+
+To make changes to the tables, edit the YAML files by hand as needed.
+Do not edit the reST tables by hand since they are generated automatically
+from the YAML files.
+
+Then, normalize the YAML files and update the reST tables:
```
jdmgoogle: How does one normalize the YAML files?

scripts/vip.py
```diff
@@ -0,0 +1,185 @@
...
+_log = logging.getLogger()
+
+FORMATTER_CLASS = argparse.RawDescriptionHelpFormatter
+
+HELP_SUMMARY = """\
+Helper script for VIP contributors.
+"""
+
+def _wrap(text):
+    """Format text for help outuput."""
```
jdmgoogle: output

jdmgoogle:
The disclaimer here is that this is a moderate size chunk of python, and python isn't my strongest language. Down the road I'd like to see some of this code replaced with something like off-the-shelf like the docutil RST tableparser:

http://svn.python.org/projects/external/docutils-0.6/docutils/parsers/rst/tableparser.py

That's not for today, though.

If @pstenbjorn or @jungshadow (or anyone else with a good Python background) could give this a thorough once-over, I'd really appreciate that.

This gets a tentative LGTM from me.

cjerdonek:
> Down the road I'd like to see some of this code replaced with something like off-the-shelf like the docutil RST tableparser:

The good news is that the table-parsing code can be removed from the repo _entirely_ after initially generating the YAML files (which will hopefully happen later today or else tomorrow).  The reason is that, after the YAML files are in place, we will only be _generating_ reST tables and not parsing them. :grinning:

cjerdonek:
PR updated with typo fix.

(Also, I forgot to mention in my previous comment that the parsing code is largely confined to `parsing.py`, which is one of the improvements I made over the previous PR.  So that file can potentially be removed in its entirety after the YAML file generation.)

jdmgoogle:
Okay. I'm all for deleting code. :)

LGTM

Cc @jungshadow @pstenbjorn @jen-tolentino

jungshadow:
@cjerdonek I have a single issue, which is the Markdown footnote in the credits. Can you put the footnote on the same line or use superscript tags (pretty sure they're allowed)? Other than the cosmetic issue, LGTM.

cjerdonek:
Okay, thanks, guys.  @jungshadow I added the superscript.  If that's not what you wanted, let me know, and I can change it in a later commit.  Or you can change it.

cjerdonek:
PS - I'm going to make a new PR right now to actually run the script, etc.

cjerdonek:
See PR #273 for part 2.

cjerdonek:
Actually, it's probably better to leave the reST parsing code in the repo for a _little_ bit longer.  It will be useful to address @jdmgoogle's suggestion to also include in the YAML the text preceding and following each table.  That would be easy to do, and would let us get rid of a bunch of the reST files (many of which will be small after issue #255).  However, it's probably better for that to be done after this week.

PR #273, opened 2015-09-02
Address issue #255: Take Two (Part 2)

cjerdonek:
This PR runs the commands added in PR #272 to get things started.

Specifically, it runs (in two separate commits):

    $ python scripts/vip.py rest2yaml
    $ python scripts/vip.py update_tables

(The `rest2yaml` command doesn't need to be run again in the future.)

I also checked that Sphinx builds okay.  There were only two errors that were left-over from before:

    tables/elements/state.rst:10: ERROR: Unknown target name: "ocd-id".
    checking consistency... docs/xml/xml_data_best_practices.rst:: WARNING: document isn't included in any toctree


jdmgoogle:
I'm going to LGTM this, and then go through and make the rest of the changes recommended in the related bugs. After that if everyone could go through and do one final pass of the documentation prior to release, that would be tremendously valuable.

jungshadow:
Agree with @jdmgoogle. LGTM.

cjerdonek:
Thanks, guys.

PR #296, opened 2015-12-14
Merge vip5 branch into master branch

cjerdonek:
Here are the steps I followed to make this PR:

First, get the vip5 branch up-to-date with changes in master by merging master into vip5:

    $ git checkout vip5
    $ git merge --strategy-option ours master
    
(The merge step turned out to be a no-op, which isn't surprising.)

Now merge the vip5 branch into master:

    $ git checkout master
    $ git merge vip5

Since this was a "fast-forward" merge, there were no conflicts.  Push this branch into my personal fork:

    $ git push


jdmgoogle:
Thank you, Chris.

I'm going to optimistically LGTM this under the assumption it won't break everything. :)

Comments, @pstenbjorn?

jungshadow:
Awesome, @cjerdonek! Weird...I could have sworn I tried that approach and ran into the next-level craziness. Maybe [this is a more accurate depiction of my git skills](http://bukk.it/howidogit.gif).

pstenbjorn:
First of all a big thank you to @cjerdonek - @jdmgoogle I will join in your optimistic LGTM on the same grounds.

cjerdonek:
Great! Thanks, everybody.

cjerdonek:
Oh, and two more things:

1) I'm not sure if the repo is set up for gh-pages to be updated automatically, or if something manually needs to be done.

2) The master branch should now be merged into the vip51 branch if that's where new development is taking place. That way vip51 will be up-to-date with master and future merges between master and vip51 will be smoother.  Should be doable with:

    $ git checkout vip51
    $ git merge master

jungshadow:
@cjerdonek Maybe this is where I'm confused, but shouldn't number 2 be a `git rebase` command to keep the history in order? https://git-scm.com/docs/git-rebase

cjerdonek:
@jungshadow No, if the commits on the branch have already been pushed for other developers to access, you wouldn't want to use rebase on them because, as you say, it would reorder the commits.  That would mean you'd need to use `git push --force` to overwrite the previously pushed branch.  And then other developers would get error messages when they next try to git fetch / pull since the branch would no longer be consistent with what they last fetched.

Also, from personal experience, git rebase doesn't really work when there is non-linear history.  The workflow assumes linear commits.

Thus, in my experience, git rebase is better for reordering your _local, linearly ordered_ changes when git fetching / pulling the latest commits from other developers.  That way other developers see a cleaner history from you instead of merge commits from your personal workflow.

jungshadow:
@cjerdonek Ah. That makes sense.

Issue #300, opened 2016-01-11
Generate the reST files from the YAML files


cjerdonek:
This is something I was supposed to do for the "5.1 Documentation" milestone, but I don't see an issue for it.  It is also a pre-requisite for issue #248.

To summarize, we're currently generating only the reST tables from the YAML files, but not the full reST files.  Generating the full files will make a lot of things simpler.  For example--

1. We'll be able to generate different versions of the docs automatically, like a "single-page" version (for issue #248) and a CSV version.  Currently, the CSV docs are a copy-paste of the XML docs with minor variations, meaning the same docs have to be maintained and updated in two different places.
2. We'll be able to in-line the tables in the reST files, instead of having to use "includes."  This will make them easier to check manually on GitHub, for example, or by viewing the reST file directly.

I prepared a PR and will submit it shortly.

cjerdonek:
I created PR #301 for this.  I can start working on issue #248 once this issue is addressed.

cjerdonek:
Note to self: we will want to use the [`:ref:` role](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-ref) to disambiguate links to sections in the various doc formats.  For example, the single-page ref name for `ExternalIdentifiers` might be `single-page-external-identifiers`, and similarly for XML and CSV (e.g. `xml-external-identifiers` and `csv-external-identifiers`).

cjerdonek:
Closed by PR #301.

PR #301, opened 2016-01-11
Issue #300: generate the reST files from the YAML files

cjerdonek:
This PR addresses issue #300.

A note on this PR: this PR only generates the reST files for the XML docs and not the CSV docs.  The CSV docs are a large copy-paste of the XML docs, so it will take a bit more work to know how they differ from the XML docs.  Adding support for the CSV docs will likely mean adding an additional field to the YAML for the CSV example, etc, and then tweaking the script that generates the reST.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,9 +74,8 @@ easy to edit by hand.  The YAML files are stored in the
 ## Dev Environment
 
 This section explains how to set up your local development environment for
-contributing.  First, [install Python][python_download].  **We recommend
-installing the latest stable version of Python 3** (Python 3.4.3 as of June
-2015).
+contributing. First, [install Python][python_download]. **You must use Python
+3.4 or higher.**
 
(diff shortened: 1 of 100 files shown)


jdmgoogle:
LGTM

Paging @pstenbjorn

pstenbjorn:
Just one minor issue in vote variation there is inconsistency of capitalization: 
``proportional representation`: https://en.wikipedia.org/wiki/Proportional_representation`
The remaining entries are first letter caps, but LGTM.

cjerdonek:
Thanks, @pstenbjorn.

Regarding the casing, that's how it was before.  (No changes were made to the content.)  The reason the casing is different is because the casing needs(?) to match how it appears in the sentence for Sphinx to tie the hyperlink correctly.  And "proportional representation" is the only one of the terms that appears in the middle of the sentence as opposed to at the beginning (hence it isn't capitalized).  Note that the casing there isn't something that is visible to the reader.  It's only reST markup for Sphinx purposes of generating the hyperlink.

Issue #248, opened 2015-06-12
Provide documentation as a single page


JohnArntz:
Hello.  As an elections official it would be most helpful to  be able to view the list of objects and fields on one screen or page without having to click through many pages and also read code (XML).  If I were to have a list of objects and fields that I can easily reference without having to know how to read code (human readable, in other words), it would better facilitate my providing pertinent information in open data formats.  Then, those who do code using this information would have more content to develop.

jdmgoogle:
@JohnArntz Hi, welcome to the VIP GitHub!

That definitely seems like a good request. @cjerdonek or @jungshadow, would we be able to script something like this?

cjerdonek:
@jdmgoogle Yeah, my sense is that this should be pretty doable.  After poking around, it might not even require any scripting (e.g. perhaps using multiple "include" directives in an `xml/single_page.rst`).  I should be able to help with this and will have more time to start taking a look next week or so.

jdmgoogle:
@cjerdonek Thanks. Assigning this one to you for now.

jdmgoogle:
Ping @cjerdonek: have you had a chance to look into this?

cjerdonek:
> Ping @cjerdonek: have you had a chance to look into this?

@jdmgoogle Just catching up on these pings.

My plan for this was to wait until PR #256 was resolved before digging in -- in case the changes involved would be conflicting.  There is a chance that this issue will require some restructuring / moving things around in some form, so I didn't want to have two such PR's in play at once.

jdmgoogle:
@cjerdonek Just looping you back in to get a sense of your current workload and availability for 5.1 stuff.

cjerdonek:
I'll have some time.  It's on my todo list in particular to work on this issue, as well as a couple of the other "build" related issues.  I expect to start working on those within the next couple weeks or so.

cjerdonek:
I didn't see an issue for it (the discussion might only have been by e-mail), but the first thing I was going to do is move the header / footer / etc stuff to the YAML like you had suggested, @jdmgoogle.

jdmgoogle:
@cjerdonek Okay, good to know. We should probably have a meeting this week just to scope out VIP 5.1 in general. If it's not this week, next week likely won't work (pre-Thanksgiving vacation), and then it's December.

cjerdonek:
[...]

PR #299, opened 2016-01-08
Change IDREF to IDREFS

jdmgoogle:
Change IDREF to IDREFS for all elements which are both IDREF and maxOccurs=unbounded.

This addresses issue #298.

Paging @cjerdonek to make sure I generated the RST from the YAML correctly.

PR #302, opened 2016-01-13
Address issue #248: expose single-page version of XML docs

cjerdonek:
Okay, I think this is good enough to merge now.  In particular, there are now fewer Sphinx warnings than there were before.  (There were lots of Sphinx warnings affecting the CSV files before I started the previous PR.)

I'm hoping this can be merged again before the "Id" / "Ids" change, because there was a fair amount to do here.  The bulk of the work, aside from updating the Python, was to replace `:doc:` references with `:ref:` references, which are preferred and more flexible for inter-document linking.  This had to be done semi-manually which would be a pain to do again.

There is one issue that can be taken care of in a subsequent commit, namely making all hyperlink "refs" choose the appropriate single-page vs. multi-page ref, depending on which version the link resides in.  That can be done dynamically / automatically by the Python when it is generating the single vs. multi-page files.


cjerdonek:
There are other improvements in this PR, too, like putting all of the auto-generated reST files in a separate `docs/built_rst` directory.  This way the auto-generated files can be kept separate and not intermingled with the reST files that need to manually edited.  It makes things clearer that way, too.

cjerdonek:
>There is one issue that can be taken care of in a subsequent commit, namely making all hyperlink "refs" choose the appropriate single-page vs. multi-page ref,

Okay, this is now done, too.  And I've also brought the number of Sphinx warnings down to zero.

jdmgoogle:
LGTM

Ping @pstenbjorn

jdmgoogle:
Also I'm fine merging these documentation changes into #299 as we go so you won't have to changes your PRs.

pstenbjorn:
@cjerdonek and @jdmgoogle this LGTM.  Thank you @cjerdonek!

cjerdonek:
Thanks a lot, @jdmgoogle and @pstenbjorn.

Issue #289, opened 2015-11-09
Support for city-level wildcarding


indymag:
In VIP 3.0, we supported city-level wildcarding (all people in city X vote at location Y). This was done by requiring a * in the street_name field of the non_house_address.

Consider supporting the ability to model these types of assignments in the VIP 5.1 spec.

jdmgoogle:
Are there any objections to this feature?

jungshadow:
@indymag @jdmgoogle Would it make sense to include another element like [IncludesAllAddresses](http://vip-specification.readthedocs.org/en/vip5/xml/elements/street_segment.html) for city-wide coverage?

jdmgoogle:
@jungshadow That's the plan. And were you bored over the break? :)

jungshadow:
@jdmgoogle What better way to digest an obscene excess of food than randomly chiming in on GitHub issues? ;)

jdmgoogle:
This has been closed by #306.

PR #306, opened 2016-01-22
Added an IncludesAllStreets to support city wildcarding.

jdmgoogle:
This should address #289.

### DIFF
--- a/docs/built_rst/tables/elements/ordered_contest.rst
+++ b/docs/built_rst/tables/elements/ordered_contest.rst
@@ -8,7 +8,7 @@
 |                           |               |              |              |                                          | ``OrderedContest`` element containing it.       |
 +---------------------------+---------------+--------------+--------------+------------------------------------------+-------------------------------------------------+
 | OrderedBallotSelectionIds | ``xs:IDREFS`` | Optional     | Single       | Links to elements that extend            | If the field is invalid or not present, the     |
-|                           |               |              |              | :ref:`multi-xml-ballot-selection-base`.  | implementation is required to ignore it. If a   |
+|                           |               |              |              | :ref:`multi-xml-ballot-selection-base`.  | implementation is required to ignore it. If an  |
 |                           |               |              |              |                                          | ``OrderedBallotSelectionIds`` element is not    |
 |                           |               |              |              |                                          | present, the presumed order of the selection    |
 |                           |               |              |              |                                          | will be the order of                            |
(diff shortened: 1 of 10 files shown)


jungshadow:
I could see an argument both ways on this, but, if `IncludesAllStreets` is "true", could the inclusion of `OddEvenBoth` potentially be error-prone? It's useful for `IncludesAllAddresses`, but I can see how it might cause problems in this case. If y'all have already discussed this and sorted it out, LGTM.

jdmgoogle:
Ping to @pstenbjorn @nomadaisy @MariaBianchi @cjerdonek.

pstenbjorn:
@jdmgoogle this looks good to me as far as documentation is concerned.  I am curious how this might be represented in a csv load.

jdmgoogle:
@pstenbjorn You're curious as in "I would like you, @jdmgoogle, to explain your thoughts" or "I would like @nomadaisy to weigh in"?

pstenbjorn:
@jdmgoogle as in, I am sure @nomadaisy has factored this in the csv, just interested in the mechanics.

jdmgoogle:
Ping @nomadaisy

jdmgoogle:
Ping @MariaBianchi

MariaBianchi:
@jdmgoogle We'll take a look at this, and though I don't have an answer at this moment, I'll try to have one by next week.  This is one of several areas where I think the CSVs need another look.

jdmgoogle:
Friendly ping to @MariaBianchi and @nomadaisy

nomadaisy:
I did include it as a boolean field in the street_segment.txt CSV structure documentation and example, so we are good to go. I envision the other address fields being blank if this is **true**, and the processor should ignore any populated address fields if this field is **true**.

jdmgoogle:
@nomadaisy Good suggestion. I updated the documentation to make `OddEvenBoth` optional. PTAL, along with @pstenbjorn and @cjerdonek

pstenbjorn:
@jdmgoogle if that is the case, the following change might also be needed:

`<xs:element name="OddEvenBoth" type="OebEnum"  minOccurs="0"/>`

jdmgoogle:
@pstenbjorn Good catch. Updated.

nomadaisy:
LGTM

jdmgoogle:
I'm going to optimistically assume that @pstenbjorn is okay with this. If he isn't, then all of you are witnesses that I'm okay with him yelling at me at the next VIP conference. :)

pstenbjorn:
Yes, so no yelling required.

Issue #268, opened 2015-08-21
Party affiliations in an Open Primary election


Josh-LACRRCC:
California uses Open Primary voting for contests other than President / County Central Committees (aka Voter Nominated). As such, LA County differentiates between formal party nominated affiliation and candidates’ self-selected party preference.
( http://www.voterguide.sos.ca.gov/en/voter-info/elections-in-california.htm )
In order to help clarify to voters these differences, there is additional text as part of the Contest and Candidate party affiliation.

Using the California Governor as example:

In a traditional Partisan format:
  * Contest: Governor
  * Candidate: Jerry Brown – Democratic / DEM

And in a Voter Nominated format:
  * Contest: Governor – Voter Nominated
  * Candidate: Jerry Brown – Party Preference: Democratic / Pref: DEM
  <br/>or shortened<br/>
  * Candidate: Jerry Brown –Pref: Democratic / VN:DEM

The “Voter Nominated” identification text on the contest can easily be handled via the ContestBase.BallotSubTitle element. The additional text that accompanies the candidate affiliation raises a question of “how best to accomplish”. My first instinct is to create a copy of the root party, adjust the element content to fit, and then treat it like any other party using ID ref’s for its linkage to a candidate.

I.E.
```xml
  <Party id="par1">
    <Abbreviation>DEM</Abbreviation>
    <Name>
      <Text language="en">Democratic</Text>
    </Name>
  </Party>

  <Party id="par1vn">
    <Abbreviation>VN:DEM</Abbreviation>
    <Name>
      <Text language="en">Pref: Democratic</Text>
    </Name>
  </Party>
```

There is no master governing list of acceptable political parties. And active political parties can vary between jurisdictions. What are thought of as “Minor” political parties are treated no different, data wise. So, the Spec already supports variation in the list of parties.

Does this seem like an acceptable tactic? Other thoughts, concerns, alternatives?

jdmgoogle:
This approach seems reasonable to me.

@jungshadow and @cjerdonek (since Chris will also have to deal with these)?

jungshadow:
:+1: This sounds fine to me.

cjerdonek:
Hmm, it sounds like the suggestion is to create a second version of each Party object: the original Party object and a VN-prefixed version of the party.

[...]

PR #308, opened 2016-02-05
Updated sample feed to provide an example of a "Top-Two" primary.

jdmgoogle:
This addresses issue #268.

Also removed all EndorsingPartyIds fields since we're changing how those work.

Pinging @cjerdonek @pstenbjorn @Josh-LACRRCC and @kennethmbennett  for review and feedback.


cjerdonek:
Thanks, @jdmgoogle. It seems like the docs should also be updated in appropriate spots, for example saying in the "Party" docs that a Party object can also be used a "dummy" party to indicate a type of primary and also to indicate the lack of a party preference (in particular, Party objects are not just in correspondence with political parties).  And perhaps also clarifying what the various "PartyId" attributes mean.

jdmgoogle:
@cjerdonek Done.

cjerdonek:
:+1:

jdmgoogle:
Ping @pstenbjorn

pstenbjorn:
@jdmgoogle looks good.

PR #310, opened 2016-02-12
Clean up documentation examples to match the spec.

jdmgoogle:
The example code in the per-element files drifted a bit from the actual spec itself, so this change brings them back in line. Given that this is a small change, unless there are objections by 2pm ET/11am PT, I'm going to merge this in.


pstenbjorn:
LGTM

jdmgoogle:
Late, but merging.

PR #313, opened 2016-04-21
Add restriction on how to correctly use HoursOpen

indymag:
It is difficult to handle overlaps in the Schedule elements.  I recommend not allowing them explicitly.  I'm guessing this was the original intention.


cjerdonek:
FYI, you probably want to be making this PR against a more recent version of the spec, like the vip52 branch. Also, in that branch you will want to make the change to the text in the YAML file rather than the XML, since the XML is auto-generated from the YAML.

jdmgoogle:
What @cjerdonek said. Please make this a PR against vip52, and see the [section of the CONTRIBUTING documentation](https://github.com/votinginfoproject/vip-specification/blob/vip52/CONTRIBUTING.md) about updating the documentation itself.

Issue #325, opened 2016-07-15
Precinct.ElectoralDistrictIds, Precinct.PrecinctSplitName, and Precinct Example Documentation Update


jungshadow:
[In the documentation](http://vip-specification.readthedocs.io/en/latest/built_rst/xml/elements/precinct.html) of `Precinct.ElectoralDistrictIds` it references `Precinct.PrecinctSplitName` and states:

> If an electoral district splits a precinct, use the PrecinctSplitName object and do not specify that particular electoral district in this object.

I thought that, from our conversations on #91, if `Precinct.PrecinctSplitName` was specified, it meant that the `Precinct` element was a precinct split, which makes this a bit confusing. After re-reading #27, it sounds like the idea is that every `Precinct` element is a precinct split unless `Precinct.PrecinctSplitName` is specified (_**NB:** how does this handle multiple splits within a precinct if `Precinct.PrecinctSplitName` can only be specified once?_) If I'm interpreting the documentation incorrectly–or my understanding of the overall concept is flawed–it would be helpful to clarify the language in those two sub-elements and add another example showing the use of `Precinct.PrecinctSplitName`.

jdmgoogle:
This is definitely confusing, thanks for pointing it out.

The basic gist is:

* If `PrecinctSplitName` is empty, then this object represents a full precinct.
* If `PrecinctSplitName` is not empty, then this represents a portion (or split) of a precinct.
* If a precinct is split into *N* parts, then there will be *N* `Precinct` objects which have the same `PrecinctName` but different `PrecinctSplitName`s.

Does this help clarify? Once we get the wording down I'll update the documentation.

jungshadow:
@jdmgoogle Ah...that makes a lot of sense. Thanks! So to make a quick-and-ugly example (_**NB:** idea is `ElectoralDistrict` elements ed60129 and ed60037 split `Precinct` pre90111_):

```xml
<Precinct id="pre90111">
   <ElectoralDistrictIds>ed60311 ed60054</ElectoralDistrictIds>
   <IsMailOnly>false</IsMailOnly>
   <LocalityId>loc70001</LocalityId>
   <Name>203 - GEORGETOWN</Name>
   <Number>0203</Number>
   <PollingLocationIds>pl81274</PollingLocationIds>
</Precinct>

<Precinct id="pre90112">
   <ElectoralDistrictIds>ed60129</ElectoralDistrictIds>
   <IsMailOnly>false</IsMailOnly>
[...]

PR #326, opened 2016-07-19
Clarify documentation surrounding precinct splits.

jdmgoogle:
This addresses #325.

Paging @jungshadow.

### DIFF
--- a/docs/yaml/elements/precinct.yaml
+++ b/docs/yaml/elements/precinct.yaml
@@ -31,9 +31,7 @@ tags:
   description: Links to an :ref:`$$$-electoral-district` (e.g., congressional district,
     state house district, school board district) to which the precinct belongs. **Highly
     Recommended** if candidate information is to be provided. Multiple allowed and
-    recommended to specify the geography of multiple electoral districts. If an electoral
-    district splits a precinct, use the `PrecinctSplitName` object and do not specify
-    that particular electoral district in this object.
+    recommended to specify the geography of multiple electoral districts.
   error_then: =must-ignore
   type: xs:IDREFS
 - _name: ExternalIdentifiers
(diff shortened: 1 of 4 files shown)


docs/yaml/elements/precinct.yaml
```diff
@@ -31,9 +31,7 @@ tags:
   description: Links to an :ref:`$$$-electoral-district` (e.g., congressional district,
     state house district, school board district) to which the precinct belongs. **Highly
     Recommended** if candidate information is to be provided. Multiple allowed and
-    recommended to specify the geography of multiple electoral districts. If an electoral
-    district splits a precinct, use the `PrecinctSplitName` object and do not specify
-    that particular electoral district in this object.
+    recommended to specify the geography of multiple electoral districts.
```
jungshadow: The confusing language is gone, but I'm wondering if you might now have an issue with folks specifying all the electoral district ids and not just the specific ones for the split. The link to the sample feed in the `Precinct.PrecinctSplitName` description might help, but I figured I'd raise the concern.

jungshadow:
Other than the one concern, LGTM 👍

docs/yaml/elements/precinct.yaml
jdmgoogle: It's actually correct behavior (I believe) to specify all the electoral district IDs for each split, and not just the ones where they differ. @indymag?

jdmgoogle:
Wait, did you just try to LGTM a PR? As a non-Pew employee??

image

jungshadow:
I am a placebo.
Dr Doom Toots as he pleases

jdmgoogle:
Ping @indymag @pstenbjorn @rhoffmanpew for a review

docs/yaml/elements/precinct.yaml
indymag: Yes, I agree.  Reviewing this now, I think the documentation should note that the electoral districts should be _wholly contained_.  Suggest: changing  "to which the precinct belongs" to be "to which the entire precinct or precinct split is contained within).

Without some note like that, I fear it could be interpreted as any intersection with the precinct/spit.

indymag:
See note about mentioning it is wholly contained.  Either incorporate in this change or roll it separately.  The change so far is LGTM.

pstenbjorn:
@jdmgoogle I agree with the clarification proposed by @indymag and aside from that LGTM

jdmgoogle:
Updated per @indymag and @pstenbjorn feedback. PTAL.

indymag:
LGTM

pstenbjorn:
LGTM

Issue #332, opened 2016-08-15
Add IsWriteIn to Party object


jdmgoogle:
Add a boolean `IsWriteIn` to the `Party` object. This is needed because some candidates are allowed to enter free-form Party names to identify themselves in some instances (looking at you, Washington and Wisconsin) and those names may not correspond to a real political party.

The rule of thumb for deciding if a Party is a write-in party or not is: is this a political party which is recognized by an election administration in this city/county/state/country? If yes, then the value should be false. If not, then this is a write-in party which may entirely be the creation of one candidate and feed consumers shouldn't attempt to ready anything else into the party name or information.

If this field is absent then it should be treated as `false`.

pstenbjorn:
@jdmgoogle sounds confusing, but I believe I followed and LGTM

jdmgoogle:
@pstenbjorn Some states (e.g., WI) allow candidates to enter the name of the party they want shown next to them on the ballot. This party name doesn't have to line up with any registered party, and can in fact be whatever (within reason) the candidate wants to have next to their name. This flag is basically a way to signal that a feed consumer shouldn't expect this political party to show up ... well ... anywhere else.

indymag:
LGTM

jdmgoogle:
Closed by #333.

PR #333, opened 2016-08-17
Add IsWriteIn to Party

jdmgoogle:
Adds a field to signal if a political party is an officially recognized one, or a 'write-in' party from a jurisdiction which allows free-form entry of political affiliation.

Addresses #332.

Paging @indymag and @pstenbjorn


indymag:
LGTM

jdmgoogle:
(There was a LGTM from @pstenbjorn via email, I believe. Merging based on that.)

Issue #336, opened 2016-08-23
Type child element for Locality should be optional


indymag:
For a Locality, the Type element is listed as optional in the documentation, but is required in the XSD.  It should probably be optional.

http://vip-specification.readthedocs.io/en/release/built_rst/xml/elements/locality.html
https://github.com/votinginfoproject/vip-specification/blob/vip52/vip_spec.xsd#L444

jdmgoogle:
@indymag Could you explain a bit more about why you think Type should be optional?

indymag:
It isn't really needed for our use case.  We are linking Locality elements to Polling Locations for early voting.  Those locality elements are referenced by external identifiers to their OCD Division ID.    Type doesn't add value for our use case.

Additionally, the documentation says its optional though.  So its an easy mistake to make.  

Is there a reason why Type was required?

jdmgoogle:
It was required in the [3.0 spec](https://github.com/votinginfoproject/vip-specification/blob/v3-archive/vip_spec_v3.0.xsd), so that probably just carried forward. If this is legitimately causing an issue then we can ditch it.

indymag:
I think we should relax for 5.2 if it isn't really needed.  In the meantime, I'll see if a workaround is feasible.

PR #337, opened 2016-08-30
Make Locality Type optional

jdmgoogle:
This addresses #336.

This -- along with the `Party.IsWriteIn` -- will need to be folded back into the next revision of the NIST 1500-100 standard.

Paging @indymag


indymag:
LGTM but I don't understand the NIST compatibility tag.  I think this is a VIP-only element.

jdmgoogle:
I'm sorry, I was thinking of the `Candidate.ContactInformation` change. Too many issues in-flight this morning.

jdmgoogle:
Paging @pstenbjorn @MariaBianchi @nomadaisy and really anyone else for the last set of eyes on this.

MariaBianchi:
Yep. LGTM.

PR #355, opened 2016-12-27
Remove old 4.0 schema, update to 5.1.2

sduckett:
When we write 5.1 XML, we begin with:
`<VipObject xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
schemaVersion="5.1"
xsi:noNamespaceSchemaLocation="http://votinginfoproject.github.com/vip-specification/vip_spec.xsd">`

This was (quietly) pointing to a 4.0 schema, but that doesn't work!


jdmgoogle:
What is referencing the gh-pages branch?

sduckett:
> What is referencing the gh-pages branch?

The [5.1 XML we write](https://github.com/votinginfoproject/data-processor/blob/master/src/vip/data_processor/output/tree_xml.clj#L95) references the `gh-pages` branch; seeing it again this morning, I'm inclined to point it to `master` instead.

jdmgoogle:
Ah, okay. In that case, yes, my recommendation is to not do that anymore. :)

PR #369, opened 2018-01-31
Bug/issue 368 contest base.electoral district

afsmythe:
ContestBase.ElectoralDistrictId was incorrectly flagged as optional. This PR corrects that, flagging the element as Required.


jdmgoogle:
LGTM. Just needs one more approver.

PR #370, opened 2018-02-02
Update CONTRIBUTING.md

lbirdpew:
1. Adding @mariabianchi as an approver to the repo
2. Editing the language in contributing under section 4 for readablity

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,14 +34,9 @@ Discussion."
     2. if working on a documentation bug/feature, the documentation must build with Sphinx with no
        errors (_**NB:** see [Installing Sphinx](#installing-sphinx) below_).
 3. Once it's done and tested, create a pull request to move it into the current working branch.
-4. At that point, some discussion might happen. In order to get approval for the pull request, you
-   will need approval from two people, including one representative from Pew and one representative
-   from Google (Pew and Google employees still need two approvers and cannot self-approve, but it
-   is not required that the second approver be from the organization of the PR author). However it
-   is important to note that other members have substantial technical and election background as
-   well, so please take all feedback to heart, regardless of the source.
+4. Some feedback might be necessary in order to approve a PR. Many members of this repository have substantial technical and election   backgrounds, so please take all feedback to heart, regardless of the source.  In order to get approval for the pull request, you will need approval from two people, including one representative from Pew and one representative from Google.  Pew and Google employees still require two approvers and cannot self-approve, but it is not required that either approver on the PR be from the same organization as the PR author. 
     1. Google approvers: @jdmgoogle
-    2. Pew approvers: @lbirdpew; @afsmythe
+    2. Pew approvers: @lbirdpew; @afsmythe; @MariaBianchi
 5. When it's reviewed and accepted by the team within a reasonable timeframe (TBD), it's merged
    into the current working branch by the developer who created the pull-request.
 6. Delete the feature/bug branch.


afsmythe:
Taking the opportunity to ask a general question about how VIP's github branches are structured and to which branches certain updates are going to need to be merged with. I am generally unfamiliar with why Read the Docs surfaces the documentation the way that it does. For instance, there are three RTD branches currently; release, latest, vip52 (the latter only added sometime during last year). I, perhaps incorrectly, have been merging documentation updates (usually to correct bad documentation) for the current spec (5.1.2) in to the vip52 branch (see [this PR](https://github.com/votinginfoproject/vip-specification/pull/369)). Coincidentally, those updates are not available in the latest and release RTD branches. Question for @jdmgoogle: where should corrections to bad documentation be merged? Having some guidelines around which branches need which PRs would be great!

Making this comment a bit more topical, should this PR 370 be made against either the release or latest branch? Assuming those branches remain available for updates to documentation, should the `Contributing.md` for those branches be updated to guide the approval process for updates there?

jdmgoogle:
Disclaimer: I'm really bad at managing GitHub branches.

The intent of each of the following branches was as such:

- **vip5X** stable branch which points to the VIP 5.X release (e.g., vip51, vip511, vip512)
- **release** stable branch which always points to the most recent released version of the spec
  + Changes would not get committed here; it would instead serve as a stable pointer for URLs
- **latest** / **master**: These branches -- **latest** on RTD, and **master** on GitHub -- would point to whatever version of the spec and documentation we're working on now. The contents of this branch would be unstable and potentially inconsistent. In other words, here there be dragons.

What actually appears to have happened is that I made **vip52** the default branch (the one which would probably better be named **latest** or **master**).

Also, I generally haven't had to deal with backports of documentation, or when I have done them I may have done them badly.

To address each question:
1. Bad documentation should be fixed in whatever branch is actually relevant. I don't know the best way to do that off the top of my head, but if someone here has experience with backporting changes into an existing branch but making sure they also propagate forward, please weigh in. Otherwise I'll eventually give it a go myself.
2. Where PR 370 should go could end up being a test case for part 1, as we figure out how to get it into the VIP 5.1.2 branch and the current (unstable/latest/master) one.

Thoughts?

cjerdonek:
Re: backporting changes vs. “propagating forwards,” generally what pretty
much all projects do I’ve seen (e.g. CPython) is to make the change in the
newest branch in which it applies, and then apply the change separately to
the older branches. The main reason for this is that it lets you prepare
the change in the go-forward version, which is worth investing the most
effort into (since it applies to all future versions).

You can do the actual backporting using “git cherry-pick.” The reason for
cherry-picking is that a plain merge would merge all changes up to that
point, which you of course wouldn’t want (since the branch is a newer
branch containing newer stuff).

There are fancier ways that take more preparation and would let you do
straight merges, but I wouldn’t recommend those because it’d make it harder
to explain and understand (and possibly make mistakes).

jdmgoogle:
Thank you, @cjerdonek.

@lbirdpew In that case let's merge this PR into vip52, and then you can cherry-pick it into the vip512 branch.

afsmythe:
In addition to `vip512` I think this should also be cherry-picked in to the `master` branch (pointing to the "latest" Read the Docs build. Is that correct?

lbirdpew:
@afsmythe can you approve the PR officially then I will merge.  Also, can someone else do the cherry pick to the other appropriate branches?

PR #371, opened 2018-02-07
Bug/issue 368 contest base.electoral district

afsmythe:
Test merge per comments in [this PR](https://github.com/votinginfoproject/vip-specification/pull/370). I don't know if this in fact a "cherry pick" but it seems this branch is okay to merge to vip512.


cjerdonek:
The difference between a "cherry-pick merge" and a "merge" is that a cherry-pick just merges a single "delta" or commit, while a normal merges an entire branch, which includes not just the latest commit but all the commits leading up to it that haven't been merged yet.

So to cherry-pick a single commit, you would do this:

    $ git cherry-pick <40-character SHA>

instead of--

    $ git merge <branch>

Judging from the changes here, it looks like you're doing a merge rather than a cherry-pick since there are so many changes.

Issue #351, opened 2016-11-15
Provide a way to provide structured addresses for Polling Locations


indymag:
We moved from structured to unstructured addresses for polling locations in the transition from 3.0 to 5.1.  I'm uncertain if this was a win.  We had numerous problems in interpreting unstructured addresses.  Additionally, it added a bit of unpredictability to the system.  It wasn't always clear to the data provider what needed to be done in order to get an address through the system.  A structured address adds a degree of predictability that shouldn't be discounted.

My preference would be to move to fully structured addresses for VIP 5.2 (don't allow unstructured).

afsmythe:
Support of unstructured polling location addresses removes the requirement to parse out the address fields in to structured columns. This on the surface would seem to simplify the data provider's work to generate a polling location list. With 5.1 the provider only needed to provide a polling location name and unstructured address to identify the polling place.

afsmythe:
For the 5.2 revision we're hoping to support structured or unstructured polling location addresses. Supporting either will allow 5.1 states to continue VIP submissions without upgrading while new upgrades to 5.2 can take advantage of a structured address type. 

From a high level, it sounds like we'll need to incorporate a "structured address" complexType in to the `PollingLocation` element and specify in the XSD that `PollingLocation` requires either the new structured address type or a value for the unstructured `PollingLocation.AddressLine`, which is presently required. If both are provided, the structured address should take precedence in implementation. 

Does this summary sound correct? And should other objects in the spec which currently use non-structured addresses (e.g. `ContactInformation`) also incorporate the new structured address complexType?

afsmythe:
Does the [3.0 `simpleAddressType`](https://github.com/votinginfoproject/vip-specification/blob/6f931f9e53725c5fa0179326552d5d75122becef/vip_spec_v3.0.xsd#L336) incorporate all the necessary elements?

jdmgoogle:
I think it does incorporate the necessary elements. I'd propose to use xs:choice in order to make sure that providers use either AddressLine or a simple address (componentized) but not both.

afsmythe:
The `xs:choice` XSD option is included for these two items in the `PollingLocation` object: https://github.com/votinginfoproject/vip-specification/blob/ba3eaa2a0dd359ea663c317f21697a3dfd29292a/vip_spec.xsd#L537

JDziurlaj:
[...]

Issue #377, opened 2019-02-07
Add an alias type field to the PollingLocation object 


afsmythe:
Many VIP states will sometimes refer to their election day polling locations, early vote sites, and drop off locations differently from how VIP labels them. Examples include Colorado having "[VSPC's](http://justvotecolorado.org/)"
>Every Coloradan has the option to vote in person rather than returning their mail ballot. You may vote in person at any Voter Service and Polling Center (VSPC) in your county. VSPCs open several weeks before Election Day. 

Georgia has "[Advanced Voting](http://elections.sos.ga.gov/Elections/advancedVotingInfo.do)" as opposed to "Early Voting". Other terms in some states include "Mail in absentee", "Drop off locations", "Drop boxes", and others. Some VIP States have complained about their not having the ability to label the VIP `PollingLocation` objects appropriately. 

A proposed solution to this is to add a new field or incorporate the existing Type/OtherType logic to the `PollingLocation` class which will allow each record to be labeled individually if desired. This field would ideally be surfaced by the Civic Info API to allow the `PollingLocation` records to be labeled accordingly to what voters in that state expect.

afsmythe:
Closing this on account of possible usage of `PollingLocation.Name` to describe the type of location, as mentioned here: https://github.com/votinginfoproject/vip-specification/pull/387

PR #334, opened 2016-08-17
Add a Name element to the PollingLocation object.

jdmgoogle:
This addresses #330.

Paging @indymag @pstenbjorn @jktomer @MariaBianchi

PR #387, opened 2019-10-23
Feature/issue 351 structured polling locations

afsmythe:
This includes updates the YAML, RST and XSD files to support structured addresses for PollingLocation objects. The 5.2 revision will give the option to provide structured or unstructured addresses for polling locations via `choice` in the XSD file. Further discussion around this revision update is found here: https://github.com/votinginfoproject/vip-specification/issues/351

This PR also relocates the PollingLocation XML example from `lat_lng.yaml` to `polling_location.yaml`.

### DIFF
--- a/docs/built_rst/tables/elements/simple_address_type.rst
+++ b/docs/built_rst/tables/elements/simple_address_type.rst
@@ -7,9 +7,9 @@
 |              |               |              |              | structured address.                      | then the implementation is required to   |
 |              |               |              |              |                                          | ignore it.                               |
 +--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
-| Line1        | ``xs:string`` | Optional     | Single       | The address line for a structured        | If no ``Line1`` is provided, the         |
+| Line1        | ``xs:string`` | **Required** | Single       | The address line for a structured        | If no ``Line1`` is provided, the         |
 |              |               |              |              | address. Should include the street       | implementation should ignore the         |
-|              |               |              |              | number, stree name, and any prefix and   | ``SimpleAddressType``.                   |
+|              |               |              |              | number, street name, and any prefix and  | ``SimpleAddressType``.                   |
 |              |               |              |              | suffix.                                  |                                          |
 +--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
 | Line2        | ``xs:string`` | Optional     | Single       | TBD                                      | If the field is invalid or not present,  |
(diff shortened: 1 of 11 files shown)


docs/yaml/elements/simple_address_type.yaml
```diff
@@ -0,0 +1,54 @@
...
+  description: The name of the building a part of the structured address.
+  error_then: =must-ignore
+  repeating: false
+  required: false
+  type: xs:string
+- _name: Line1
+  csv-header-name: structured_line_1
+  csv-type: xs:string
+  description: The address line for a structured address. Should include the street
+    number, stree name, and any prefix and suffix.
```
jdmgoogle: stree name? :)

afsmythe: Thanks. Fixed :)

afsmythe:
@jdmgoogle apologies but by accident I merged in the commits to add a [`PollingLocation.Alias`](https://github.com/votinginfoproject/vip-specification/issues/377) to this PR. Hope thats okay but if you'd rather they be separate PRs I can try to fix.

jdmgoogle:
I'm not 100% clear on what an alias is, or how it's different from the name. The description makes it sound like some class of polling place, not an alias. What type of information, specifically, is trying to be captured here? What's the motivation for adding it?

afsmythe:
Thanks, @jdmgoogle. I think you're right in categorizing this as a class of PollingLocation. 

We've gotten some feedback from States who would prefer to be able to use their own terms when describing a polling location. For example, in Colorado they use "VSPC" instead of "Polling Location" and would like that information surfaced in the API in some way. In Georgia, early voting locations are referred to as "Advanced Voting Sites" rather than "Early Voting Sites". Hope that adds some clarity and I'd be happy to provide a bit more detail. 

Is this a change that would be better implemented with Enumerations? If yes, I think we could put together a list of what the States are using.

jdmgoogle:
I think that's something that belongs in the `Name` field.

Also, how many of these changes are because that's how the law/election officials refer to them, versus how the general public refers to them? E.g., if we pull Random J. Voter off the street in Georgia and ask them if they plan to go to a local precinct or early voting site, will we get a blank look, a long pause and a ".... early voting? You mean ... Advanced Voting Site?" or will they know what we mean?

afsmythe:
Most 5.1 states are using `PollingLocation.Name` to list the building name of the polling location. [This PR](https://github.com/votinginfoproject/vip-specification/pull/334) also seems to agree that name of the building, rather than a description of the location type, is what should be placed here. 

The reason for adding an Alias type filed to `PollingLocation` is because this has been requested by the elections offices in the States. As far as we know, the elections offices use these terms (VSPC in Colorado,  Vote Center in many states, etc.) and so I think it is likely that a learned voter in the State would know what is meant by such a field.

jdmgoogle:
I understand that the states are asking for this in the spec, but this would require work to plumb through end-to-end across all surfaces. Additionally, I'm trying to figure out the benefit to the user since these are already getting tagged in the schema as early voting locations. Is there actual user benefit to this, or is this purely because the states want to see words that are familiar to them?

vip_spec.xsd
```diff
@@ -531,7 +531,10 @@
   <xs:complexType name="PollingLocation">
     <xs:sequence>
       <!-- It is intentional that an address is required. -->
-      <xs:element name="AddressLine" type="xs:string" maxOccurs="unbounded" />
+      <xs:choice>
+        <xs:element name="AddressStructured" type="SimpleAddressType"/>
```
jdmgoogle: While I *think* this will work, could you try and do XML validation on an old 5.1 file using this proposed XSD?

afsmythe: The `sample_feed.xml` in this PR does validate against the XSD and there is 1 PollingLocation with a AddressStructured in that file.

afsmythe: There are also PollingLocation records with unstructured addresses in sample_feed.xml, so I think we're good.

docs/built_rst/tables/elements/simple_address_type.rst
```diff
@@ -0,0 +1,22 @@
+.. This file is auto-generated.  Do not edit it by hand!
+
++--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| Tag          | Data Type     | Required?    | Repeats?     | Description                              | Error Handling                           |
++==============+===============+==============+==============+==========================================+==========================================+
+| Line1        | ``xs:string`` | **Required** | Single       | The address line for a structured        | If no ``Line1`` is provided, the         |
```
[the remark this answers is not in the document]
afsmythe: We can add line2 and line3. These were available in the v3 spec.

docs/built_rst/tables/elements/simple_address_type.rst
```diff
@@ -0,0 +1,22 @@
...
+|              |               |              |              |                                          | implementation should ignore the         |
+|              |               |              |              |                                          | ``SimpleAddressType``.                   |
++--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| State        | ``xs:string`` | **Required** | Single       | The State value of a structured address. | If ``State`` is not provided, the        |
+|              |               |              |              |                                          | implementation should ignore the         |
+|              |               |              |              |                                          | ``SimpleAddressType``.                   |
++--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| Zip          | ``xs:string`` | **Required** | Single       | The ZIP code of a structured address.    | If ``Zip`` is not provided, the          |
```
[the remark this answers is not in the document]
afsmythe: ZIP was required in the v3 spec, along with City and State. But I think ZIP could be optional for our intents for 5.2. Will a data consumer (e.g. Google) be able to validate a geocoded address using only the City and State fields?

Issue #367, opened 2018-01-17
Add field to ElectionAdministration object to associate ballot tracking URL


afsmythe:
Presently, the `ElectionAdministration` object provides for URL's of various purposes to be associated with the administration of that district, e.g. Absentee voting (AbsenteeUri), registration (AmIRegisteredUri), elections (ElectionsUri) etc. Some elections offices have additional tools made available to mail-in voters allowing the voter to track their absentee ballot as it moves through the postal service, i.e. [Denver's Ballot Trace](https://www.denvergov.org/content/denvergov/en/denver-elections-divison/voter-election-information/ballot-trace.html), among others. Adding a field to the VIP specification (say, BallotTrackUri) to handle this specific type of URL would further make the 5.1 specification attractive to states providing such tools. 

P.S. pinging @MariaBianchi - feel free to add additional info here.

jungshadow:
@afsmythe I may be missing something, but I'm a bit confused here. Are you referring to the [`ElectionAdministration` element](https://vip-specification.readthedocs.io/en/release/built_rst/xml/elements/election_administration.html) (not `ElectoralDistrict`)?

afsmythe:
Woops. Yes, @jungshadow. Thanks for catching! I've updated the original description.

jungshadow:
Awesome! Not that it matters, but sounds like a good idea to me. In general (and I know it's not easy), it would be great if the existing links were more consistently populated.

jdmgoogle:
Do we have any idea:

1. How many states or counties have something like this in place?
2. How many people use such a feature?
3. Where this data would get surfaced?

I support the idea and it makes sense for the spec. The question is if this should go into a 5.1.x release (maybe ready in time for this cycle) or a larger 5.2 cleanup for 2020.

jungshadow:
@jdmgoogle A few thoughts to your questions.

> How many states or counties have something like this in place?

Though I'm not sure of the overall number, quite a few states have some way for absentee and UOCAVA voters to track their ballots.

> How many people use such a feature?

Unsure of the numbers. I believe there are some studies that mention absentee voters feel more secure knowing that a method to track their ballot exists (even if they don't use it). @afsmythe, y'all have one. Do you have anything you can share?

> Where this data would get surfaced?

I'd surface it on all appropriate election admin elements. Since ballots go out for UOCAVA voters much earlier, it's useful info for a good number of people, even if polling locations and ballot data aren't available, yet.

[...]

PR #389, opened 2019-10-24
Feature/issue 367 add ballot tracking uris

afsmythe:
This will allow VIP participating States and Counties to list web urls in the `ElectionAdministration` object to point users to where they may find information about the tracking of mail-in ballots and provisional ballots with the latter being a suggestion from the EAC. 

Discussion thread here: https://github.com/votinginfoproject/vip-specification/issues/367


jdmgoogle:
Merging.

jdmgoogle:
Actually, no.

I still need someone else to take a look at this. I believe @sarahberlin is the other approver right now.

PR #403, opened 2020-04-20
Feature/issue 400 add locality ismailonly

afsmythe:
We're adding an `IsMailOnly` flag to the Locality element to efficiently flag that a Locality maintains vote by mail for each of it's precincts.


afsmythe:
Pinging for review.

afsmythe:
Thanks! Merging..

PR #418, opened 2021-07-21
Replaces deprecated Sphinx.add_stylesheet with Sphinx.add_css_file

jswiesner:
Sphinx.add_stylesheet was deprecated with version 1.80. See https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -372,4 +372,4 @@ epub_exclude_files = ['search.html']
 
 # Inject a custom stylesheet into the Read the Docs theme
 def setup(app):
-  app.add_stylesheet("theme_overrides.css")
+  app.add_css_file("theme_overrides.css")


afsmythe:
Thanks for looking in to this. I'll note that I haven't used Sphinx for some time, opting instead for Python's http server offering. Would it help to also document that option in addition to Sphinx?

jswiesner:
Is the idea that we'd move doc generation entirely off of Sphinx? If not, I'm pretty sure this change still needs to be made in order to build the HTML documentation.

afsmythe:
>Is the idea that we'd move doc generation entirely off of Sphinx?

No. Your change is needed, since it appears readthedocs uses Sphinx to build the documentation. Though, if there was an error raised during documentation build, I didn't see it when `vip6` was surfaced initially.

afsmythe (approved):

jswiesner:
My guess is that you have an older version of Sphinx that hasn't fully turned down that method yet. I just installed it for the first time today using pip so I likely have the latest and greatest version.

PR #422, opened 2021-08-10
Updates geospatial data best practices doc based on feedback

jswiesner:
Includes the following changes:
-  Adds a section to clarify that VIP exclusively uses the WGS 84 datum
- Clarifies that requirements for precinct shapes apply to the "Precinct" element, not a literal precinct. This should help clarify that these requirements apply to precincts as well as precinct splits, whichever the Precinct element represents.
- Adds detail around the use case of multiple polygons to define a single Precinct boundary.

### DIFF
--- a/docs/data_best_practices.rst
+++ b/docs/data_best_practices.rst
@@ -203,6 +203,16 @@ the boundary of voter precincts. In places where voter precinct shapes
 are available, this capability is intended to be straightforward and
 lightweight to integrate with existing GIS tooling.
 
+Geodetic Datum
+~~~~~~~~~~~~~~
+
+VIP exclusively uses the 84 revision of the World Geodetic System (WGS
+84) as the geodetic reference system by which geospatial coordinates
+are defined. This applies to geospatial coordinates provided within the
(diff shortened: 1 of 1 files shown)


docs/data_best_practices.rst
```diff
@@ -203,6 +203,16 @@ the boundary of voter precincts. In places where voter precinct shapes
...
 lightweight to integrate with existing GIS tooling.
 
+Geodetic Datum
+~~~~~~~~~~~~~~
+
+VIP exclusively uses the 84 revision of the World Geodetic System (WGS
+84) as the geodetic reference system by which geospatial coordinates
+are defined. This applies to geospatial coordinates provided within the
+VIP feed itself (e.g. PollingLocation.LatLng) as well as coordinates
+provides in an external geospatial file.
```
afsmythe, line 213: typo: provides > provided

afsmythe (changes requested):
Minor typo otherwise looks good, thanks!

docs/data_best_practices.rst
jswiesner: Good catch, fixed.

afsmythe (approved):
LG!

PR #424, opened 2021-11-04
Vip6 feature identifier

afsmythe:
Addition of a `FeatureIdentifier` type to `ExternalGeospatialFeature`. 

This will move the feature index (previously `ShapeIdentifier`) to a `FeatureIdentifier` complex type, housing an `Index` to store the index value of the geospatial feature.


afsmythe:
`FeatureIdentifier` is a repeating field so that multiple shapes can be assigned to a `Precinct`. Should `FeatureIdentifier` be encapsulated within a `FeatureIdentifiers` element? Or is it suitable that repeating `FeatureIdentifier` elements be nested directly underneath `ExternalGeospatialFeature`? E.g.

```
          <ExternalGeospatialFeature>
            <ExternalFileId>ef1</ExternalFileId>
            <FileFormat>shp</FileFormat>
            <FeatureIndentifier>
                <Index>3</Index>
            </FeatureIdentifier>
            <FeatureIdentifier>
                 <Index>4</Index>
            </FeatureIdentifier>
          </ExternalGeospatialFeature>
```
or
```
          <ExternalGeospatialFeature>
            <ExternalFileId>ef1</ExternalFileId>
            <FileFormat>shp</FileFormat>
            <FeatureIndentifiers>
                <FeatureIndentifier>
                  <Index>3</Index>
                </FeatureIdentifier>
                <FeatureIdentifier>
                   <Index>4</Index>
                </FeatureIdentifier>
            </FeatureIdentifiers>
          </ExternalGeospatialFeature>
```

docs/built_rst/tables/elements/feature_identifier.rst
```diff
@@ -0,0 +1,10 @@
+.. This file is auto-generated.  Do not edit it by hand!
+
++--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| Tag          | Data Type     | Required?    | Repeats?     | Description                              | Error Handling                           |
++==============+===============+==============+==============+==========================================+==========================================+
+| Index        | ``xs:string`` | **Required** | Single       | The index value for the shapefile        | If the field is invalid, then the        |
```
jswiesner, line 6: WDYT about making this optional? In the future I could imagine a FeatureIdentifier element containing various different ways to refer to a shape (e.g. attribute-based identifier). In that case, it would be reasonable to provide the index or an attribute, in which case both would need to be optional in the spec to allow that flexibility.

docs/built_rst/xml/elements/precinct.rst
```diff
@@ -104,7 +104,9 @@ is assigned to that precinct.
         <ExternalGeospatialFeature>
           <ExternalFileId>ef1</ExternalFileId>
           <FileFormat>shp</FileFormat>
-          <ShapeIdentifier>3</ShapeIdentifier>
+          <FeatureIndentifier>
```
jswiesner, line 107: typo: FeatureIndentifier

docs/built_rst/xml/elements/precinct.rst
```diff
@@ -142,14 +144,14 @@ SpatialBoundary
...
-|                           |                                              |              |              |                                          | it.                                      |
-+---------------------------+----------------------------------------------+--------------+--------------+------------------------------------------+------------------------------------------+
++----------------------------+----------------------------------------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| Tag                        | Data Type                                    | Required?    | Repeats?     | Description                              | Error Handling                           |
++============================+==============================================+==============+==============+==========================================+==========================================+
+| ExternalGeospatialFeatures | :ref:`multi-xml-external-geospatial-feature` | **Required** | Single       | The spatial boundary defined by a        | If the element is invalid, then the      |
```
jswiesner, line 150: Is this intentional to pluralize ExternalGeospatialFeature?

jswiesner:
> Should `FeatureIdentifier` be encapsulated within a `FeatureIdentifiers` element? Or is it suitable that repeating `FeatureIdentifier` elements be nested directly underneath `ExternalGeospatialFeature`?

I'm open to other thoughts, but I personally prefer the approach of 1 or more FeatureIdentifier elements nested directly underneath ExternalGeospatialFeature.

jswiesner (changes requested):

docs/built_rst/tables/elements/feature_identifier.rst
afsmythe: Sounds good to me.

docs/yaml/elements/feature_identifier.yaml
```diff
@@ -0,0 +1,10 @@
+_name: FeatureIdentifier
+csv-header-name: feature_identifier
+tags:
+- _name: Index
+  csv-header-name: index
+  csv-type: xs:string
```
jswiesner, line 6: WDYT about making the index field an integer? When this used to be called "ShapeIdentifier", I originally chose a string format to offer greater flexibility for other shapefile formats. But since we plan to extend the FeatureIdentifier approach using other additional fields within the FeatureIdentifier element in the future (e.g. add an attribute field), we'll get flexibility that way instead. So this seems reasonable to me to be an integer field.

jswiesner (changes requested):

docs/built_rst/tables/elements/feature_identifier.rst
JDziurlaj: Also what if the dereferenced `ExternalFile` only represents a single polygon? Should `FeatureIdentifier` be required at all?

jswiesner: Yes, FeatureIdentifier should still be required in that case. The ExternalGeospatialFeature is a reference to a single polygon within an external file containing one or more polygons. The FeatureIdentifier is how we know which single polygon to grab from that file. 

It's worth nothing that even if we make FeatureIdentifier.Index optional, one or more resolvable values is still required in a FeatureIdentifier element. This is a constraint that will need to be enforced in code to check 1) there is a usable value in each FeatureIdentifier and 2) each successfully resolves to one or more polygons in an external file.

docs/yaml/elements/feature_identifier.yaml
afsmythe: SGTM! Thanks. Update forthcoming.

docs/built_rst/csv/elements/precinct.rst
```diff
@@ -127,33 +127,47 @@ external_geospatial_feature
...
+
+.. _multi-csv-feature-identifier:
+
+feature_identifier
+^^^^^^^^^^^^^^^^^^
+
++--------------+---------------+--------------+--------------+------------------------------------------+------------------------------------------+
+| Tag          | Data Type     | Required?    | Repeats?     | Description                              | Error Handling                           |
++==============+===============+==============+==============+==========================================+==========================================+
+| index        | ``xs:string`` | Optional     | Single       | The index value for the shapefile        | If the field is invalid or not present,  |
```
jswiesner, line 170: Looks like the CSV docs still need to be updated, the index field is still showing as a string type.

jswiesner (approved):
This LGTM, but I think the CSV rst still needs to be re-generated. Approving assuming the CSV docs will be fixed before merging. Thank you for putting this together!

PR #427, opened 2021-12-01
Replacing v5 sample with v6 sample

afsmythe:
Replacing v5 sample with v6 sample.


jswiesner (commented):
It looks like the new sample feed doesn't have any street segments. Would it be worth keeping the existing sample feed and possibly renaming it to indicate the version? I could imagine the segment-based sample feed may still be useful for data providers that don't have shapes it. One possible naming scheme would be to rename the existing feed to sample_feed_v5.xml and the new sample feed is sample_feed_v6.zip.

jswiesner (approved):
This looks great! I'm very much looking forward to testing out these precinct shapes!

PR #433, opened 2021-12-14
Feature/issue 432

afsmythe:
Changing _ExternalFile.FileName_ from `xs:string` to `xs:anyURI`.


jswiesner (commented):
Two questions:
1. Will the XSD change be in a separate PR?

2. Should we rename this field from Filename to FileUri? It looks like we always use the Uri suffix for fields in the XSD of type xs:anyURI. If we did, we'd also have to update the v6 sample feed.

afsmythe:
XSD update will be a part of this PR. Sorry about that and thanks for catching. 

I think it makes sense to rename `Filename` to `FileUri`. Update to this PR (including sample feed) will follow. 

If we expect to be comfortable supporting a linked dataset via URL, then we'll need to update the [Best Practices](https://github.com/votinginfoproject/vip-specification/blob/vip6/docs/data_best_practices.rst#exporting-and-packaging-geospatial-files-with-a-vip-feed). But I don't see us wanting to do that for the pilot. 

Remaining tasks for this PR:
- XSD update for FileUri field and type
- sample feed update to rename `Filename` to `FileUri`.

jswiesner:
SGTM!

docs/yaml/elements/external_file.yaml
```diff
@@ -16,15 +16,15 @@ post: |-
...
           <Value>65b634c5037f8a344616020d8060d233daa37b0f032a71d0d15ad7a5d3afa68e</Value>
         </Checksum>
      </State>
 tags:
-- _name: Filename
-  csv-header-name: filename
+- _name: FileUri
+  csv-header-name: file_uri
   csv-type: xs:anyURI
   description: The name of the external file.
```
jswiesner, line 29: We should update the description to say "URI" instead of "name".

jswiesner (approved):

Issue #279, opened 2015-09-04
Improve comments in python YAML code


jdmgoogle:
Right now the python code that parses the YAML documentation and generates the RST pages is lightly commented, which makes maintaining it more difficult. This needs to be improved.

afsmythe:
With the 2020 primaries, we didn't get to this. Punting to 5.3 and hopefully we can comment the functions sooner than later.

jswiesner:
Considering we will all be on a trajectory of increasing busyness between now and 2024, I propose we move this up to the 6.0 milestone and take care of it this year.

afsmythe:
Agreed.

jswiesner:
IMO [vip.py](https://github.com/votinginfoproject/vip-specification/blob/vip52/scripts/vip.py) looks sufficiently well commented, [common.py](https://github.com/votinginfoproject/vip-specification/blob/vip52/scripts/vippy/common.py) isn't bad, and [rest.py](https://github.com/votinginfoproject/vip-specification/blob/vip52/scripts/vippy/rest.py) could probably use more comments. 

@afsmythe are you able to take a first pass at adding comments to this file? LMK if you think vip.py or common.py could use more comments as well, I could take a pass at those if needed.

afsmythe:
Yes I can make a pass at documenting.

jswiesner:
@afsmythe have you had a chance to take a pass at documenting? I believe this is the last step before we are ready to release 6.0.

afsmythe:
Working on this.. Happy to contribute to a PR is something is already in progress. Should have something by first of next week.

jswiesner:
Great, thank you! I haven't done any work on this yet, so AFAIK your changes would be the first. I'm happy to help out should there be any remaining gaps after your initial pass.

jswiesner:
@afsmythe I think we can close this now?

afsmythe:
Yes! Closed.

PR #436, opened 2021-12-15
Fixes the rest.py script to properly order pages in the single-page doc

jswiesner:
The current single page doc (e.g. [the XML page](https://vip-specification.readthedocs.io/en/vip6/built_rst/xml/single_page.html)) is out of order, and the order is not deterministic. This change ensures pages are ordered before they are added to the single-page doc. 

This also updates the `update_rest_files` method to combine the outer loops over mode (xml, csv). I believe the previous implementation could yield unexpected results since `type_names` and `type_map` were assembled in the first loop and used in the second.

### DIFF
--- a/scripts/vippy/common.py
+++ b/scripts/vippy/common.py
@@ -119,8 +119,9 @@ def configure_yaml():
 
 
 def read_yaml(path):
+    #https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation
     with open(path) as f:
-        data = yaml.load(f)
+        data = yaml.load(f, Loader=yaml.FullLoader)
     return data
 
 
--- a/scripts/vippy/rest.py
+++ b/scripts/vippy/rest.py
@@ -253,7 +253,8 @@ def update_rest_file_single_page(all_types,mode):
         rest += "\n\n"
         label = "{0}-{1}".format(prefix, title.lower())
         rest += make_rest_header(title, label=label, header_char="-")
-        for data_type in data_types:
+        sorted_data_types = sorted(data_types, key=lambda dt: dt.csv_name if mode == "csv" else dt.name)
+        for data_type in sorted_data_types:
             if data_type.is_sub_type and data_type.is_extends:
                 continue
             new_rest = make_type_rest(all_types, data_type, header_char="~", prefix=prefix)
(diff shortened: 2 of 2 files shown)


afsmythe:
Thanks for catching this! Rendered documentation looks good locally, but I needed to bring in bdf68e7d23c324e3f670f05815ee5b3ca69dbfa7 in order to successfully run `python scripts/vip.py make_rest` 

We'll want to get that commit to `vip52` and `vip6` but this PR looks good for `master`.

afsmythe (approved):

afsmythe (changes requested):
Correction: I think we'll need bdf68e7d23c324e3f670f05815ee5b3ca69dbfa7 merged to `master` here as well.

jswiesner:
Sounds good, I cherry-picked bdf68e7d23c324e3f670f05815ee5b3ca69dbfa7 and included it in this PR.

jswiesner:
While testing regenerating the docs using these changes to the Python script, I noticed the table files weren't being generated right (it was generating these files under the case of `mode` = "csv", whereas the current table files in the repo were generated with `mode` = "xml"). I'm not sure how it worked in the past since we always process modes in the order of ("xml", "csv"), but I updated the call to `update_table_file` so we only do this step once using the XML mode.

afsmythe:
Sorry, is there any additional information you could add about the "table files"? I'm not seeing any suspicious formatting on rendered documentation built locally.

jswiesner:
By "table files" I meant the generated files under [built_rst/tables](https://github.com/votinginfoproject/vip-specification/tree/vip6/docs/built_rst/tables). We generate these files with calls to `update_table_file`, which we currently do twice - once for mode = "xml" and again for mode = "csv". If you comment out the if statement I added around the call to `update_table_file` and then regenerate the RST files, you'll notice that the files in built_rst/tables are updated to use the csv field names and prefix. I assumed the desired behavior here was to preserve the XML field names, hence I added the if statement to be compatible with the existing files there.

But now that I look at this more closely, I can't find any pages that references the files in built_rst/tables. Are you aware of anything that uses these? If not, we could just delete these files and the code that generates them.

jswiesner:
I also found this, which looks like we explicitly exclude files from the tables directory when generating the docs: https://github.com/votinginfoproject/vip-specification/blob/c694e79d40b6b71ec9ae94c7cda5a916915c75fb/docs/conf.py#L84

afsmythe:
Ok, interesting.. this may be a vestige of when the documentation only supported the XML format. CSV was added later and I'm wondering if the `tables` resource was retained when it could have been removed. Will investigate, but we could probably handle this in another PR if you want to go ahead and merge your code related to the simple page ordering.

afsmythe:
Noting the `tables` question might be better handled by the forthcoming PR related to https://github.com/votinginfoproject/vip-specification/issues/279 

Happy to take that on within that PR.

jswiesner:
That sounds great! I also filed #438 so we can separately track the clean up of the tables directory if we do end up confirming it's not needed any more. 

I'm going to go ahead and merge this. I'll send follow-up PRs to merge this change into the vip52 and release branches as well, and then also regenerate the docs using the correct ordering.

PR #439, opened 2021-12-16
Regenerates all doc RST to align with the latest script version

jswiesner:
This PR is the outcome of running the following. The main change here is that the single page RSTs now contain all the elements in lexicographical order.

```
$ python3 scripts/vip.py norm_yaml
$ python3 scripts/vip.py make_rest
```


afsmythe (approved):
LG!

Issue #438, opened 2021-12-16
Possibly delete built_rst/tables from the docs


jswiesner:
It looks like the RST files in [built_rst/tables](https://github.com/votinginfoproject/vip-specification/tree/master/docs/built_rst/tables) are generated by the Python script, but aren't referenced anywhere. We might be able to delete this directory altogether, as well as the Python code that generates them.

jswiesner:
Hi @cjerdonek. Since you wrote a lot of the original doc generation code, I wanted to ask if you happen to recall if [built_rst/tables](https://github.com/votinginfoproject/vip-specification/tree/master/docs/built_rst/tables) is still in use for anything? We believe these directories are obsolete, but would be great to get your input before we delete them.

PR #444, opened 2022-01-05
Feature/issues 279 438

afsmythe:
To resolve:
https://github.com/votinginfoproject/vip-specification/issues/279
https://github.com/votinginfoproject/vip-specification/issues/438

Note that `vippy.rest.get_next_header_char` needs some additional input.


scripts/vippy/rest.py
```diff
@@ -288,6 +322,9 @@ def update_rest_files(type_name=None):
```
jswiesner, line 322: We'll need to remove calls to update_table_file since the method is removed with this PR. Should we go ahead and remove all the table generation methods as well while we're at it (e.g. make_table)?

afsmythe: I believe all calls (there was 1) to `update_table_file` were removed in https://github.com/votinginfoproject/vip-specification/pull/444/commits/d57021123e3a82b94b041e488b5a4d8b79916d00

WRT `make_table`, this exists as a function and a method of `TableFormatter` class. The function calls the method, and AFAICT these "tables" created by function and method are the RST tables used by both csv and xml pages. So I think they should remain, but would appreciate another pair of eyes.

jswiesner: After looking again, I agree that make_table is still needed as part of the call to update_rest_file.

I'm a bit confused by what I'm seeing in this PR though for update_table_file. I see that line 286 is removed in d57021123e3a82b94b041e488b5a4d8b79916d00, and that d57021123e3a82b94b041e488b5a4d8b79916d00 is part of this PR. But when I look at the full set of changes in the PR across all 4 commits, it doesn't show line 286 as being removed. And I don't see another commit where it's added back in. Do you know why that is?

afsmythe: You're right WRT `update_table_file`. I see what happened (stale local branch which omits https://github.com/votinginfoproject/vip-specification/pull/436). Going to close this PR and track in https://github.com/votinginfoproject/vip-specification/pull/445/, which was cut fresh from `vip6`.

PR #445, opened 2022-01-13
Feature/issues 279 438 v2

afsmythe:
There is some confusion around https://github.com/votinginfoproject/vip-specification/pull/444. 

This is a second pass, cherry picking: 
54912ac5d8727408d300c9d52eb92be0b101b9f8
d57021123e3a82b94b041e488b5a4d8b79916d00


jswiesner (approved):
This LGTM. Thanks for putting this together!

afsmythe:
There seems to be an issue with the current `make_rest` procedure in vip6 where the "xml/elements" RST files are not being created properly. 

To test, check out both `vip52` and `vip6` branches, delete the contents of `built_rst` and then execute
```
python scripts/vip.py make_rest
```
For vip52, all expected folders and files will be created, but for vip6 the "xml/elements" folder is not created. 

I've tried to look through the code and am unable to see why the XML element write step is being skipped. I suspect this was introduced in https://github.com/votinginfoproject/vip-specification/commit/a1d546bba67b563af9e05e3db86f4055a5c8965d

@jswiesner can you have a look, please?

jswiesner:
Good catch, you were right! I found the issue and sent over #446 to address it. In short the argument into the method `update_rest_files` was named `type_name`, which was being reused within the method itself. The bug had always been there, but the refactoring a1d546bba67b563af9e05e3db86f4055a5c8965d exposed the issue to start causing a problem. Once that PR is merged you can cherry-pick into this PR and it should work properly.

jswiesner (approved):
LGTM!

PR #446, opened 2022-01-13
Fixes a bug in rest file generation

jswiesner:
A bug was recently introduced with a1d546bba67b563af9e05e3db86f4055a5c8965d where the name of the argument `type_name` of the method `update_rest_files` was reused in the method itself. This caused unexpected behavior when the loop over modes was executed the second time.  This bug is addressed by changing the name of the argument to this method from `type_name` to `override_type_name`, since the argument is used when the user wants to manually override which single type to process.


afsmythe (approved):
LG, thanks!

PR #451, opened 2022-01-18
Removes obsolete call to update_table_file

jswiesner:
This addresses an anomaly introduces during the merge of `vip6` branch into `master`. The intention is that this call to update_table_file should be removed (see b8179d37d99a1bab4a8afb76fce568998db6e642).


afsmythe:
Should this be merged first to `vip6` branch, which will then be merged to `master`?

afsmythe (approved):
Approving. We can handle vip6 separately.

PR #462, opened 2024-05-14
Version in the source element should be optional.

mpechter:


docs/yaml/elements/source.yaml
```diff
@@ -79,5 +79,5 @@ tags:
   csv-header-name: version
   csv-type: xs:string
   description: Specifies the version of the data
-  required: true
+  error_then: =must-ignore
```
danny-pease, line 82: (question, non-blocking)

what does this do?

danny-pease (approved):
📝

PR #466, opened 2024-05-22
Skipping tags based on mode.

mpechter:
Also, lots of black reformatting. One function renamed for clarity.

### DIFF
--- a/scripts/vippy/rest.py
+++ b/scripts/vippy/rest.py
@@ -5,10 +5,17 @@ from pprint import pformat, pprint
 import textwrap
 
 from vippy import common
-from vippy.common import (TAG_KEY_NAME, TAG_KEY_TYPE, TAG_KEY_REQUIRED,
-                          TAG_KEY_REPEATING, TAG_KEY_DESCRIPTION,
-                          TAG_KEY_ERROR_HANDLE, TAG_KEY_EXTENDS, TAG_KEY_CSV_TYPE,
-                          TAG_KEY_CSV_HEADER_NAME)
+from vippy.common import (
+    TAG_KEY_NAME,
+    TAG_KEY_TYPE,
+    TAG_KEY_REQUIRED,
+    TAG_KEY_REPEATING,
+    TAG_KEY_DESCRIPTION,
+    TAG_KEY_ERROR_HANDLE,
+    TAG_KEY_EXTENDS,
+    TAG_KEY_CSV_TYPE,
+    TAG_KEY_CSV_HEADER_NAME,
+)
 
 
 _log = logging.getLogger()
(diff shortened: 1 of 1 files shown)


scripts/vippy/rest.py
```diff
@@ -92,21 +99,26 @@ def make_table_formatter(all_types, data_type, prefix):
...
+        widths=widths,
+        cell_values=cell_values,
+        prefix=prefix,
+    )
     return formatter
 
 
-
-def make_table(all_types, data_type, prefix):
+def make_standalone_table(all_types, data_type, prefix, mode):
```
mpechter, line 113: Previously there had been two functions named `make_table()`; this one and a method of a class.

scripts/vippy/rest.py
```diff
@@ -164,16 +178,11 @@ def get_next_header_char(char):
...
-        "~": "^",
-        "^": "%",
-        "%": "^"
-    }
+    mapping = {"=": "-", "-": "~", "~": "^", "^": "%", "%": "^"}
     return mapping[char]
 
-def make_type_rest(all_types, data_type, header_char, prefix):
+
+def make_type_rest(all_types, data_type, header_char, prefix, mode):
```
mpechter, line 185: Passing mode for the skip.

scripts/vippy/rest.py
```diff
@@ -455,8 +471,10 @@ def make_table(self, data_type):
             tags = data_type.tags
             widths = self.make_widths(self.headers, tags)
             lines = [self.make_divider(widths)]
-            lines.extend(self.make_row(self.headers, widths, separator='='))
+            lines.extend(self.make_row(self.headers, widths, separator="="))
             for tag in tags:
+                if mode == tag.get("skip_on", ""):
```
mpechter, line 476: Here is the crucial logic: if the tag contains a line saying "skip_on: csv", will not incorporate the tag into that RST file. This allows us to have some elements (e.g. parent_id) that are exclusive to either CSV or XML.

danny-pease: suggtestion, non-blocking

is "skip_on" a name that we choose? i did some googling and such and i think it is, but i may be wrong. if it is a name we choose, how would you feel about using a logically positive identifier rather than a logically negative one? that is to say, instead of `skip_on: xml` we could have `relevant_specification` which could be "csv", "xml", or "both". i find that somewhat easier to read - but that's just me!

am I overthinking this? i might be overthinking this. and it's subjective after all, so feel free to disregard. plus theres the non-trivial downside of needing to add this flag to everything - but i kind of consider that a positive in the long run. food for thought!

danny-pease (approved):
LGTM! had a suggestion but this is good to go either way

scripts/vippy/rest.py
mpechter: I will think about this. I agree about reading affirmative vs negatives.

PR #468, opened 2024-05-30
adding ability to skip whole element by mode.

mpechter:


scripts/vippy/common.py
```diff
@@ -352,36 +354,41 @@ def __init__(self, yaml, data, rel_path, snake_name, is_enum):
...
-        return self.data['csv-header-name']
+        return self.data["csv-header-name"]
 
     @property
     def csv_type(self):
-        return self.data['csv-type']
+        return self.data["csv-type"]
+
+    @property
+    def skip_element_on(self):
```
mpechter, line 374: This grabs the new "tag" or label from the YAML, incorporates it as an attribute of this object.

scripts/vippy/rest.py
```diff
@@ -305,6 +305,8 @@ def update_rest_file_single_page(all_types, mode):
             data_types, key=lambda dt: dt.csv_name if mode == "csv" else dt.name
         )
         for data_type in sorted_data_types:
+            if mode == data_type.skip_element_on:
```
mpechter, line 308: This is the crucial logic, skipping the element entirely.

scripts/vippy/rest.py
```diff
@@ -329,8 +331,10 @@ def update_rest_files(override_type_name=None):
 
         prefix = "multi-{}".format(mode)
         for type_name in type_names:
-            _log.debug("updating rest files for type: {0}".format(type_name))
             data_type = common.get_type(type_map, type_name)
+            if mode == data_type.skip_element_on:
```
mpechter, line 335: And the same here, skipping when creating the individual (as opposed to the single-page) rst file.

danny-pease:
@mpechter is this a repeat of the PR from the other day? or am i having deja vu

mpechter:
Great question! That was a PR to skip a specific field. This is for skipping an entire Element.

For example, we want to skip HoursOpen for CSV. Then we can configure a separate, CSV-only Schedule entity.

scripts/vippy/common.py
danny-pease: question non-blocking

by "skip element on", do you mean "on" as opposed to "off"? or is "on" intended to imply an event (as in, for example, [onmouseover](https://www.w3schools.com/jsref/event_onmouseover.asp))

if the latter, which event? and if the former, is "skip element" ever set to off? and if so, where?

danny-pease (approved):
LGTM! i added one question that isn't blocking

btw, if you wanted to push up a PR that applies the linter to the whole repo (so that future changesets are more obvious to reviewers) i would happily smash the approve button!

scripts/vippy/common.py
mpechter: It is the latter. The "event" is the run mode, where the YAML is being used to produce either of the CSV or XML rst files. @danny-pease

PR #486, opened 2025-01-16
Simplifying how error handling is explained.

mpechter:
I have stripped out the complex logic that handles error policies. In its place is a system that allows for custom error statements for complex cases, but otherwise allows the "Required" or "Optional" status of a field speak for itself.

This simplifies the code dramatically, but is also makes the exceptional cases more clear and makes the documentation less cluttered.


docs/yaml/elements/street_segment.yaml
```diff
@@ -89,6 +86,7 @@ tags:
     the even side, or both are in included in the street segment.
   error: If the field is not present or invalid, the implementation is required to
     ignore the StreetSegment containing it.
+  required: true
```
mpechter, line 89: I believe this to be correct.

docs/yaml/elements/term.yaml
```diff
@@ -21,18 +21,15 @@ tags:
   csv-type: OfficeTermType
   description: Specifies the type of office term (see :ref:`$$$-office-term-type`
     for valid values).
-  error: If the field is invalid or not present, the implementation is required to
-    ignore the ``Office`` element containing it.
+  required: true
```
mpechter, line 24: I believe I'm interpreting this correctly.

YAML_README.md
```diff
@@ -201,26 +201,16 @@ The type describes the data type for the tag. Can indicate a sub-element (in XML
 If a tag has `required: true`, the `Required` field will reflect that.
```
mpechter, line 201: In the past, apropos of this error handling issue, I've mentioned the desire to add "Conditionally Required". However, I think the format expressed here is more clear. An exceptional case is explained in the "Error Behavior" value, as needed. This is very rare.

benedictbrown (approved):

YAML_README.md
mpechter: @jennifertlovell

jennifertlovell (approved):
Your interpretation makes sense to me. Excited to see how neat the docs look after this.

PR #489, opened 2025-05-15
remove department, move contactInfo to electionAdmin

jones-liam-j:
In my opinion this is one of the most confusing and underutilized parts of the specification.

In 6.0 and earlier, the `Department` object only exists to hold contact information and/or a PersonId for the election official (which might be redundant as `Person` elements can have their own contact information).

Meanwhile the `VoterService` object is a child of the `Department` and holds its own contact information as well as a description of the provided service (voter registration, overseas voting, etc.)

This proposal removes `Department` entirely and makes `VoterService` a direct child of `ElectionAdministration`, as well as giving `ElectionAdministration` its own contact information field.

Strictly speaking this could lead to data loss from older versions of the feed, but as I said these elements are very underutilized and I believe any loss would be minimal.


benedictbrown:
Let's discuss this when we meet.  It looks like a good change that will actually bring VIP's ElectionAdministration entity more in line with CDF's equivalent entity, which would be a win.

benedictbrown (approved):

PR #490, opened 2025-05-28
Restore Source element

jones-liam-j:
This element was erroneously removed from the XSD in [commit 2e40d54](https://github.com/votinginfoproject/vip-specification/commit/2e40d54b310be056e2ddb72de154c8ba60558d99).

### DIFF
--- a/vip_spec.xsd
+++ b/vip_spec.xsd
@@ -713,6 +713,21 @@ and will enforce that each text language is different.
     <xs:attribute name="language" type="xs:language" default="i-default" />
   </xs:complexType>
 
+  <xs:element name="Source" minOccurs="1" maxOccurs="1">
+    <xs:complexType>
+      <xs:all>
+        <xs:element name="DateTime" type="xs:dateTime" />
+        <xs:element name="Description" type="InternationalizedText" minOccurs="0" />
+        <xs:element name="FeedContactInformation" type="ContactInformation" minOccurs="0" />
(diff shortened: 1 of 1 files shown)


vip_spec.xsd
```diff
@@ -713,6 +713,21 @@ and will enforce that each text language is different.
...
+  <xs:element name="Source" minOccurs="1" maxOccurs="1">
+    <xs:complexType>
+      <xs:all>
+        <xs:element name="DateTime" type="xs:dateTime" />
+        <xs:element name="Description" type="InternationalizedText" minOccurs="0" />
+        <xs:element name="FeedContactInformation" type="ContactInformation" minOccurs="0" />
+        <xs:element name="Name" type="xs:string" />
+        <xs:element name="OrganizationUri" type="xs:anyURI" minOccurs="0" />
+        <xs:element name="TermsOfUseUri" type="xs:anyURI" minOccurs="0" />
+        <xs:element name="VipId" type="xs:string" />
```
jones-liam-j, line 725: @benedictbrown To my knowledge none of these fields are surfaced by the API, they only exist to document the feed file's provenance. I leave it to you whether the URIs or any other fields need to be internationalized.

benedictbrown: Source hasn't actually disappeared.  It has just moved into the VipObject definition inside vip_mainfeed_transform.xml since it isn't used anywhere else.  The idea is that this keeps it out of the overlay spec.

PR #491, opened 2025-06-11
move FileFormat from EGF to ExternalFile

jones-liam-j:
This has always confused me. A single ExternalFile can't have multiple file formats, right?


benedictbrown:
It looks like it is structured the way it is because the FileFormat is
specifically a geospatial file format; ExternalFile could refer to other
things as well in theory (even though the spec doesn't support any other
types of external files currently).  The GeospatialFormat enumeration is,
in turn, shared with CDF, so I don't think we want to generalize it.

